PR #23, opened 2015-01-28
Bugfix filepaths on windows

### LINKED ISSUE #22: Windows: use of double backslash in payload's path results in 404
Hi,
On my Windows computer, when the application tries to perform any operation it is always met with a 404 error because the payload looks like this` {"asset":{"key":"assets\ \image.jpg"}}` (added space Github trying to escape it) which is not acceptable. it has to be a single foward slash {"asset":{"key":"assets/image.jpg"}} . Is this an issue with just my computer or Windows in general?

csaunders:
Fixes #22

FYI @wolverex

### DIFF
--- a/cmd/theme-manipulate/main.go
+++ b/cmd/theme-manipulate/main.go
@@ -5,6 +5,7 @@ import (
 	"fmt"
 	"github.com/csaunders/phoenix"
 	"os"
+	"path/filepath"
 	"strings"
 )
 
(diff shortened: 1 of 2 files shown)


csaunders:
Ugh... the work I transferred over from my windows machine got all messed up. I'm going to come back at this from another angle.

Should have another PR up in like 30 minutes.

PR #34, opened 2015-02-26
Feature/bootstrap

csaunders:
Adds the `bootstrap` command based on the atom feed pulled down from Github.

### DIFF
--- a/atom.go
+++ b/atom.go
@@ -1,0 +1,50 @@
+package phoenix
+
+import (
+	"encoding/xml"
+	"io"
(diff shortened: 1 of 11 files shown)


csaunders:
Github doesn't keep caches of the `master.zip` for a really long time, so if we weren't able to create the zip we retry another couple of times before dying.

fbv3ak12z7qp

In `mitmproxy` this is what the requests look like:
g10ngvtzm5uq

And this is the error that is being returned by Shopify:
rpez8w23zqxc

PR #35, opened 2015-02-26
Use a byte buffer for uploads

### LINKED ISSUE #31: Theme watch not uploading entire file
This is an interesting one, perhaps tied to the size of the file being uploaded. When using `theme watch` to handle uploading [a stylesheet](https://github.com/Shopify/Timber/blob/master/assets/timber.scss.liquid) it seems to upload different amounts, my tests show it tends to drop out between lines 567–1700 (approx 45kb)

```css
.pagination-custom-lg > li:first-child > a,
.pagination-custom-lg > li:first-child > span {
  border-bottom-left-radius: 6px;
  border-top-left-ra
```

No such errors when using `theme upload`

csaunders:
Fixes #31

/cc @thisislawatts

### DIFF
--- a/theme_client.go
+++ b/theme_client.go
@@ -1,6 +1,7 @@
 package phoenix
 
 import (
+	"bytes"
 	"crypto/tls"
 	"encoding/base64"
 	"encoding/json"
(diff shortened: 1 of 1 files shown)


thisislawatts:
👌

PR #50, opened 2015-05-11
Bugfix theme upload

csaunders:
This fixes upload such that a request to upload will upload all files instead of having to specify the exact files.

- Also fixes a bug with git repositories
- Makes filters work within the theme client so filtering should come for free now

### DIFF
--- a/cmd/theme/main.go
+++ b/cmd/theme/main.go
@@ -16,6 +16,7 @@ const commandDefault string = "download [<file> ...]"
 var globalEventLog chan themekit.ThemeEvent
 
 var permittedZeroArgCommands = map[string]bool{
+	"upload":   true,
 	"download": true,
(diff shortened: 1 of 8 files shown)


bman-80stees:
Will this be available for windows x64 download? I'm not sure how to apply a patch to the current version I have.

csaunders:
Just replace the binary.

bman-80stees:
I had downloaded the file from the downloads area and had to append  the .exe to the end. I'm not that familiar with git and binaries.

bman-80stees:
I found where I needed to be, thanks again for these fixes!

csaunders:
I just fixed the build script so future releases shouldn't have that missing `.exe` file extension anymore.

bman-80stees:
Thanks!

Issue #65, opened 2015-09-02
Automatic Updating of theme binary


csaunders:
Getting updates is a bit of a pain and it would be nice if people could just have the application update itself instead of needing to re-install another binary.

Some options
- https://github.com/sanbornm/go-selfupdate
- https://equinox.io/

csaunders:
Related to #15

PR #84, opened 2015-09-11
Adds Release script to enable automatic updating

csaunders:
In order to perform automatic updates we need to know
where the new binary is going to be. While it is possible
to include that in the binary itself, that's fragile.

This is a basic script that builds provides two feeds
along with directories containing the new binary for
each platform.

Clients can either pull down changes from latest.json
or if the version difference is too large apply changes
one at a time (more important if we move over to binary
delta patches).

Right now we'll just upload new uncompressed binaries to
S3 each time and add a new record to the all.json feed.

There is also some guard code in the release script to
ensure that we don't attempt to create an update if
the build hasn't been run or if we've already uploaded
that version.

---------------------

I know that this doesn't have any tests, please forgive me. I don't know if it's _really_ worth testing or not. There is a fair amount of functionality in there though, so maybe I should.

Please review @nickhoffman @jamesmacaulay 

This is one step towards resolving https://github.com/Shopify/themekit/issues/65

### DIFF
--- a/release
+++ b/release
@@ -3,17 +3,16 @@ require 'fog'
 require 'dotenv'
 require 'json'
 require 'digest'
-require 'set'
 require 'git'
 
 Dotenv.load
 
-ALL_RELEASES = "releases/all.json"
-LATEST_RELEASE = "releases/latest.json"
-GIT_DIR = File.expand_path(__FILE__ + "/..")
-
 class FileManager
+  ALL_RELEASES = "releases/all.json"
+  LATEST_RELEASE = "releases/latest.json"
+
   attr_reader :bucket
+
   def initialize(bucket)
     @bucket = bucket
   end
@@ -28,16 +27,22 @@ class FileManager
   end
 
   def url(filename)
-    bucket.files.get(filename).public_url
+    get_file(filename).public_url
   end
 
   def fetch(filename, default: "")
-    return bucket.files.get(filename).body if bucket.files.head(filename)
-    return default
+    bucket.files.head(filename) ? get_file(filename).body : default
+  end
+
+  private
+  def get_file(name)
+    bucket.files.get(name)
   end
 end
 
 class Repository
+  GIT_DIR = File.expand_path(__FILE__ + "/..")
+
   def initialize
     @git = Git.open(GIT_DIR)
   end
@@ -48,25 +53,17 @@ class Repository
 end
 
 class Release
-  attr_reader :platform, :file, :version
+  attr_reader :platform, :file, :version, :full_name, :data
   attr_accessor :location
-  def initialize(version, platform, filename)
+
+  def initialize(version: nil, platform: nil, filename: nil)
     @platform = platform
     @version = version
-    @file = File.open(filename, 'rb')
-  end
-
-  def full_name
-    [version, platform, File.basename(file.path)].join('/')
-  end
-
-  def data
-    file.rewind
-    file.read
+    load_file(filename)
   end
 
   def hexdigest
-    Digest::MD5.hexdigest(data)
+    @hexdigest ||= Digest::MD5.hexdigest(data)
   end
 
   def as_hash
@@ -76,10 +73,19 @@ class Release
       "digest" => hexdigest
     }
   end
+
+  private
+  def load_file(filename)
+    file = File.open(filename, 'rb')
+    @full_name = [version, platform, File.basename(file.path)].join('/')
+    @data = file.read
+    file.close
+  end
 end
 
 class ReleaseGenerator
   attr_reader :version, :releases, :storage_manager
+
   DIST_DIR = File.expand_path(__FILE__ + "/../build/dist")
   BUILDS = %w(darwin-amd64 linux-386 linux-amd64 windows-386 windows-amd64)
 
@@ -98,6 +104,8 @@ class ReleaseGenerator
     end
   end
 
+  private
+
   def ensure_builds_have_been_created
     return if File.exist?(DIST_DIR)
     puts "Distribution build at #{DIST_DIR} has not been created. Run 'make dist' before attempting to create a new release"
@@ -105,7 +113,7 @@ class ReleaseGenerator
   end
 
   def ensure_release_is_necessary
-    feed = JSON.parse(storage_manager.fetch(ALL_RELEASES, default:  '[]'))
+    feed = JSON.parse(storage_manager.fetch(FileManager::ALL_RELEASES, default:  [].to_json))
     if feed.find { |r| r['version'] == version }
       puts "v#{version} has already been deployed. If this was intended to be a new release, ensure version.go has been updated and add an appropriate tag to git"
       exit(1)
@@ -116,13 +124,14 @@ class ReleaseGenerator
     BUILDS.map do |platform|
       build_location = File.expand_path(DIST_DIR + "/#{platform}")
       binary = Dir.entries(build_location).find { |name| name =~ /theme/ }
-      Release.new(version, platform, [build_location, binary].join('/'))
+      Release.new(version: version, platform: platform, filename: [build_location, binary].join('/'))
     end
   end
 end
 
 class FeedGenerator
   attr_reader :version, :releases, :storage_manager
+
   def initialize(version, releases, storage_manager)
     @version = version
     @releases = releases
@@ -130,20 +139,20 @@ class FeedGenerator
   end
 
   def upload!
-    storage_manager.upload!(LATEST_RELEASE, latest_feed)
-    storage_manager.upload!(ALL_RELEASES, entire_feed)
+    storage_manager.upload!(FileManager::LATEST_RELEASE, latest_feed)
+    storage_manager.upload!(FileManager::ALL_RELEASES, entire_feed)
   end
+
+  private
 
   def latest_feed
     as_hash.to_json
   end
 
   def entire_feed
-    full_feed = JSON.parse(storage_manager.fetch(ALL_RELEASES, default: "[]"))
+    full_feed = JSON.parse(storage_manager.fetch(FileManager::ALL_RELEASES, default: [].to_json))
     full_feed << as_hash
-    full_feed.each_with_object(Set.new) do |release, memo|
-      memo << release
-    end.to_a.to_json
+    full_feed.uniq.to_json
   end
 
   def as_hash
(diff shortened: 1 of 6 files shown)


release
```diff
@@ -0,0 +1,175 @@
...
+    file = bucket.files.new(
+      key: filename,
+      body: content,
+      public: true
+    )
+    file.save
+  end
+
+  def url(filename)
+    bucket.files.get(filename).public_url
```
nickhoffman: `bucket.files.get(filename)` could be moved into its own method to DRY up this class a little.

release
```diff
@@ -0,0 +1,175 @@
...
+  end
+
+  def fetch(filename, default: "")
+    return bucket.files.get(filename).body if bucket.files.head(filename)
+    return default
```
nickhoffman: Ternary instead of two `return`s?

release
```diff
@@ -0,0 +1,175 @@
...
+  end
+end
+
+class Repository
+  def initialize
+    @git = Git.open(GIT_DIR)
```
nickhoffman: `GIT_DIR` should be in this class instead of polluting the global namespace.

release
```diff
@@ -0,0 +1,175 @@
...
+
+Dotenv.load
+
+ALL_RELEASES = "releases/all.json"
+LATEST_RELEASE = "releases/latest.json"
+GIT_DIR = File.expand_path(__FILE__ + "/..")
+
+class FileManager
+  attr_reader :bucket
+  def initialize(bucket)
```
nickhoffman: :cop: whitespace above this line, plz

release
```diff
@@ -0,0 +1,175 @@
...
+
+  def latest_version
+    @git.tags.last.name
+  end
+end
+
+class Release
+  attr_reader :platform, :file, :version
+  attr_accessor :location
+  def initialize(version, platform, filename)
```
nickhoffman: :cop: whitespace above this line, plz

nickhoffman: It'd be nice to use keyword args here.

release
```diff
@@ -0,0 +1,175 @@
...
+    @platform = platform
+    @version = version
+    @file = File.open(filename, 'rb')
```
nickhoffman: This FD is never closed :cry:

release
```diff
@@ -0,0 +1,175 @@
...
+  end
+
+  def full_name
+    [version, platform, File.basename(file.path)].join('/')
+  end
+
+  def data
+    file.rewind
+    file.read
```
nickhoffman: Should the result be memoized, or do we want to re-read on each method call?

release
```diff
@@ -0,0 +1,175 @@
...
+      "name" => platform,
+      "url" => location,
+      "digest" => hexdigest
+    }
+  end
+end
+
+class ReleaseGenerator
+  attr_reader :version, :releases, :storage_manager
+  DIST_DIR = File.expand_path(__FILE__ + "/../build/dist")
```
nickhoffman: :cop: whitespace above this line, plz

release
```diff
@@ -0,0 +1,175 @@
...
+  end
+
+  def ensure_builds_have_been_created
+    return if File.exist?(DIST_DIR)
+    puts "Distribution build at #{DIST_DIR} has not been created. Run 'make dist' before attempting to create a new release"
+    exit(1)
+  end
+
+  def ensure_release_is_necessary
+    feed = JSON.parse(storage_manager.fetch(ALL_RELEASES, default:  '[]'))
```
nickhoffman: Is `'[]'` supposed to be JSON? If so, `[].to_json` represents the intent better.

release
```diff
@@ -0,0 +1,175 @@
...
+    if feed.find { |r| r['version'] == version }
+      puts "v#{version} has already been deployed. If this was intended to be a new release, ensure version.go has been updated and add an appropriate tag to git"
+      exit(1)
+    end
+  end
+
+  def prepare_releases
+    BUILDS.map do |platform|
+      build_location = File.expand_path(DIST_DIR + "/#{platform}")
+      binary = Dir.entries(build_location).find { |name| name =~ /theme/ }
```
nickhoffman: Is this regex specific enough? I can't tell what files will be in `build_location` , but it might be good to include a file extension in the regex.

release
```diff
@@ -0,0 +1,175 @@
...
+      Release.new(version, platform, [build_location, binary].join('/'))
+    end
+  end
+end
+
+class FeedGenerator
+  attr_reader :version, :releases, :storage_manager
+  def initialize(version, releases, storage_manager)
```
nickhoffman: :cop: whitespace above this line, plz

release
```diff
@@ -0,0 +1,175 @@
...
+    storage_manager.upload!(LATEST_RELEASE, latest_feed)
+    storage_manager.upload!(ALL_RELEASES, entire_feed)
+  end
+
+  def latest_feed
+    as_hash.to_json
+  end
+
+  def entire_feed
+    full_feed = JSON.parse(storage_manager.fetch(ALL_RELEASES, default: "[]"))
```
nickhoffman: Same question here re: `"[]"` representing JSON.

release
```diff
@@ -0,0 +1,175 @@
...
+    full_feed << as_hash
+    full_feed.each_with_object(Set.new) do |release, memo|
+      memo << release
+    end.to_a.to_json
```
nickhoffman: Can this be simplified to `full_feed.uniq.to_json` ?

release
```diff
@@ -0,0 +1,175 @@
...
+connection = Fog::Storage.new(
+  provider: 'AWS',
+  aws_access_key_id: ENV['AWS_KEY_ID'],
+  aws_secret_access_key: ENV['AWS_SECRET_KEY']
+)
+
+bucket = connection.directories.create(
+  key: ENV['AWS_BUCKET_NAME'],
+  public: true
+)
```
nickhoffman: What happens if the directory already exists?

release
```diff
@@ -0,0 +1,175 @@
...
+require 'json'
+require 'digest'
+require 'set'
+require 'git'
+
+Dotenv.load
+
+ALL_RELEASES = "releases/all.json"
+LATEST_RELEASE = "releases/latest.json"
+GIT_DIR = File.expand_path(__FILE__ + "/..")
```
nickhoffman: It's kind of smelly to pollute the global namespace with these constants.

release
```diff
@@ -0,0 +1,175 @@
...
+
+class FeedGenerator
+  attr_reader :version, :releases, :storage_manager
+  def initialize(version, releases, storage_manager)
+    @version = version
+    @releases = releases
+    @storage_manager = storage_manager
+  end
+
+  def upload!
```
nickhoffman: It looks like this and `#initialize` should be the only public methods.

release
```diff
@@ -0,0 +1,175 @@
...
+  BUILDS = %w(darwin-amd64 linux-386 linux-amd64 windows-386 windows-amd64)
+
+  def initialize(version, storage_manager)
+    ensure_builds_have_been_created
+    @version = version
+    @releases = prepare_releases
+    @storage_manager = storage_manager
+  end
+
+  def upload!
```
nickhoffman: It looks like this and `#initialize` should be the only public methods.

release
csaunders: Good call :)

release
csaunders: I control the builds, so it should be specific enough. Unix-like builds are just `theme` meanwhile Windows builds are `theme.exe`

release
```diff
@@ -0,0 +1,175 @@
...
+      public: true
+    )
+    file.save
+  end
+
+  def url(filename)
+    bucket.files.get(filename).public_url
+  end
+
+  def fetch(filename, default: "")
```
nickhoffman: We could move `ALL_RELEASES` and `LATEST_RELEASE` into this class if we provide a public method for getting each of these files.

release
csaunders: I think it's just a no-op on Amazon if the bucket already exists. I was running the code a lot during development and I never had an issue come up because of that.

release
csaunders: Just dump them in the classes I guess?

release
nickhoffman: :+1:

release
nickhoffman: Yeah. [This suggestion](https://github.com/Shopify/themekit/pull/84#discussion_r39419982) will come in handy to avoid duplicating them in multiple classes.

csaunders:
:point_left:

release
```diff
@@ -0,0 +1,193 @@
...
+    bucket.files.head(filename) ? get_file(filename).body : default
+  end
+
+  def all_releases
+    ALL_RELEASES
+  end
+
+  def latest_release
+    LATEST_RELEASE
+  end
```
nickhoffman: These two methods aren't really needed. Their return values never change, and other classes can refer to the constants.

release
```diff
@@ -0,0 +1,193 @@
...
+      "digest" => hexdigest
+    }
+  end
+
+  private
+  def load_file(filename)
+    file = File.open(filename, 'rb')
+    @full_name = [version, platform, File.basename(file.path)].join('/')
+    @data = file.read
+    file.close
```
nickhoffman: :tada:

release
nickhoffman: :+1:

nickhoffman:
One last minor suggestion. Otherwise, it looks great.

hedgehog

release
csaunders: Yeah I wasn't super sure what you meant by your other thing.

I was thinking of doing `storage_manager.all_releases` to get the file, and `storage_manager.all_releases=` to assign it, but that gets kinda weird and I wouldn't _really_ expect an assignment to also result in an external call.

dalehamel:
Why is this going to S3 instead of a github release? This is a public repo, and github releases are stored on S3 anyways

csaunders:
I'm waiting to get into the beta for equinox.io, so I don't actually see this as something that will be around for a long time. I'd be curious to know how much the costs are for this to be a major issue.

release
nickhoffman: Sorry, I should've been more specific in that other comment. I was suggesting this:

```ruby
class FileManager
  def all_releases
    fetch(ALL_RELEASES, ...)
  end

  def latest_release
    fetch(LATEST_RELEASE, ...)
  end

  private

  def fetch(...)
    # ...
  end
end

class FeedGenerator
  def entire_feed
    full_feed = JSON.parse(storage_manager.all_releases)
    # ...
  end
end
```

PR #86, opened 2015-09-14
Ignores files that are considered to be compiled

### LINKED ISSUE #41: theme download includes liquid and non-liquid versions of assets
I used the theme download (using phoenix v0.2.0) for a theme that has .js.liquid assets, and the end result is both a .js and .js.liquid version of the asset are downloaded.

For example, if you have a file called assets/site.js.liquid, the download ends up grabbing assets/site.js and assets/site.js.liquid.

I would expect only the .js.liquid files to be downloaded.

csaunders:
This is a pretty rough estimation as to which files are
compiled, but most of them typically follow a similar scheme
like "filename.js.liquid" and so on.

There is a chance this might cause some other kinds of bugs though,
so perhaps this isn't the right approach. Another alternative is
to simply detect those files and add them to the ignore section
of the environments configuration.

Fixes #41

Please provide feedback, review, etc. @chrisbutcher

### DIFF
--- a/asset.go
+++ b/asset.go
@@ -24,6 +24,21 @@ func (a Asset) IsValid() bool {
 	return len(a.Key) > 0 && (len(a.Value) > 0 || len(a.Attachment) > 0)
 }
 
+// Implementing sort.Interface
+type ByAsset []Asset
+
+func (assets ByAsset) Len() int {
+	return len(assets)
+}
+
+func (assets ByAsset) Swap(i, j int) {
+	assets[i], assets[j] = assets[j], assets[i]
+}
+
+func (assets ByAsset) Less(i, j int) bool {
+	return assets[i].Key < assets[j].Key
+}
+
 func LoadAsset(root, filename string) (asset Asset, err error) {
 	path := toSlash(fmt.Sprintf("%s/%s", root, filename))
 	file, err := os.Open(path)
--- a/asset_test.go
+++ b/asset_test.go
@@ -7,6 +7,7 @@ import (
 	"log"
 	"os"
 	"path/filepath"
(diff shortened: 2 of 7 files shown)


chrisbutcher:
This looks good overall, but one question: is there a danger that the slice of assets might ever have a compiled and raw asset in reversed order?
http://play.golang.org/p/83LD0ws4lg

chrisbutcher:
Also, maybe you can create a Map of asset keys to dupe counts, to do a quick check as to whether the asset key exists twice? To break the tie, can you then filter assets by their filename length? i.e. "foo.js.liquid" vs "foo.js" (the latter being the compiled asset).

csaunders:
yeah the ordering is a bit of a concern. Right now Shopify is returning the
assets in order but I am pretty sure there is no promise that is always
true. I was thinking I could also sort them, so I know for sure they are
always in the right order and then follow it up with the filtering.

On Mon, Sep 14, 2015 at 3:44 PM, christopher butcher <
notifications@github.com> wrote:

> Also, maybe you can create a Map of asset keys to dupe counts, to do a
> quick check as to whether the asset key exists twice? To break the tie, can
> you then filter assets by their filename length? i.e. "foo.js.liquid" vs
> "foo.js" (the latter being the compiled asset).
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/Shopify/themekit/pull/86#issuecomment-140186737>.
>



-- 
Chris Saunders // Developer @ Shopify

chrisbutcher:
:cool: I'd suggest a change that prevents that order dependency issue, and otherwise looking good!

Issue #64, opened 2015-09-02
Theme Replace does not upload all files and provides no feedback if there was a failure


csaunders:

csaunders:
@stevebosworth, do a pull from master and see if this is still the case. There's a bunch of changes that have gone in that might fix up a bunch of issues.

Let me know if anything is still not working.

stevebosworth:
Thanks @csaunders!!!

stevebosworth:
Seems like this is still an issue.  I did a pull from master and tried `theme upload` and got the same error with `invalid argument` from one of the locales files

csaunders:
I'll need to see your configuration and usage to figure out what is going
wrong.

On Tue, Sep 8, 2015 at 1:32 PM, Steve Bosworth <notifications@github.com>
wrote:

> Seems like this is still an issue. I did a pull from master and tried theme
> upload and got the same error with invalid argument from one of the
> locales files
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/Shopify/themekit/issues/64#issuecomment-138642720>.
>



-- 
Chris Saunders // Developer @ Shopify

PR #90, opened 2015-09-15
Simplifies and fixes replacing a themes assets

csaunders:
Theme replace didn't actually do it's job correctly. It would just replace existing assets on Shopify and get sad if they did not exist. Instead we actually go out and do a correct replacement buy queueing all the remote assets to be destroyed, then replacing any duplicates with update events.

The advantage is somewhat twofold:

- Reduces overall API calls since we only delete assets that don't need to be there
- Is correct

Please review @chrisbutcher 

Fixes https://github.com/Shopify/themekit/issues/64

/cc @stevebosworth


stevebosworth:
popkey

Issue #63, opened 2015-09-02
Theme Upload freezes part-way through an entire theme upload


csaunders:
Typically it will crash with a cryptic error that doesn't give you any information as to what is wrong.

`theme upload` should be enough to replicate the issue.

csaunders:
It appears like node modules are getting caught up in here and causing things to crap out. The filters don't appear to be ignoring these.

PR #91, opened 2015-09-15
Fixes File Descriptor exhaustion

csaunders:
Files were not being closed after they were used.
Typically this didn't really cause any issues, but sometimes
when working on large projects or when running the application
for an extended period of time the app would crash with a
really undescriptive error message.

The underlying issue was because we were keeping files open.

By defering Close on the file we ensure that we don't exhaust our
File Descriptor limits.

Fixes https://github.com/Shopify/themekit/issues/63

Please review @chrisbutcher 

/cc @stevebosworth

### DIFF
--- a/asset.go
+++ b/asset.go
@@ -46,6 +46,7 @@ func LoadAsset(root, filename string) (asset Asset, err error) {
 	if err != nil {
 		return
 	}
+	defer file.Close()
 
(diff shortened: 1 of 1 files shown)


chrisbutcher:
:heart: `defer` :ship:

Issue #15, opened 2014-12-11
Package management


reggi:
Hey! Just stumbled upon this repo from [shopify_theme](https://github.com/Shopify/shopify_theme/issues/121). I know this project is very fresh, and is still being developed. I was surprised to see you have everything in downloadable releases. I'd love it if you guys used a package manager. I really don't care what package manager you guys choose, just so that I can download this dependency and embed it into my build process.

csaunders:
Installation of themekit on [Linux and OSX is now basically jokes](http://themekit.cat/install/). And when #94 gets merged updating to newer versions will be a single command.

**Install for first time:**
`curl https://raw.githubusercontent.com/Shopify/themekit/installers/install | python`

**Update Application (once v0.3.0 goes out)**
`theme update`

PR #94, opened 2015-09-18
Feature automatic updater

csaunders:
This integrates `go-update` into the project so we can make it easier for users to upgrade from one version to the next.

Previously it required that when a new update came out the developer had to download a zip and overwrite their previous copy with a new one. This was error prone and made it hard to ensure users were always on the latest version.

With the `release` script in master, as a part of the release process we can upload the binaries to S3. With that functionality there we can now leverage that for future releases.

This also includes a version bump to `0.3.0` because of the newly added feature.

Fixes https://github.com/Shopify/themekit/issues/65
Fixes https://github.com/Shopify/themekit/issues/15

Please review @chrisbutcher


csaunders:
Here is what the update required banner looks like:

screenshot 2015-09-18 09 44 21

chrisbutcher:
One question at the outset: Do we need a test for update.go?

csaunders:
I haven't done any testing for the `themekit/commands` package yet.

I know I should, but it's all CLI stuff.

I know... it's no excuse :(

chrisbutcher:
No worries! Just making sure it's not an unintentional oversight! :smile:

chrisbutcher:
LGTM!

Issue #89, opened 2015-09-15
invalid CLI Args should fail gracefully


jtgrenz:
Right now entering invalid arguments causes the tool to throw a panic when there should ideally be some error handling 

[![alt](https://screenshot.click/15-46-ptv25-6jnwm.jpg)](https://screenshot.click/15-46-ptv25-6jnwm.jpg)

is a little scary looking. A better message would be something like
```
Unknown argument "ddfgdg". 
<output usage banner>
```

csaunders:
What version of the tool are you using (`theme version`). On v0.2.7 I see some 406 errors pop up, but the application still runs correctly.

If `config.yml` is missing I get a gross error, but it's different than this one.

jtgrenz:
```
➜  ~  theme version
Theme Kit v0.2.7
```

Its actually giving me the 406 errors now too. Using the wrong config.yml gives me issue https://github.com/Shopify/themekit/issues/87 

Running the command sans config.yml file gives me a different error (probably the one you got too)

```
➜  Simple git:(master) ✗ ls
assets         config         config.yml_gem config.yml_kit layout         locales        snippets       templates
➜  Simple git:(master) ✗ theme replace kdhfkdfdf
panic: runtime error: invalid memory address or nil pointer dereference
[signal 0xb code=0x1 addr=0x10 pc=0xe1e72]

goroutine 10 [running]:
net/http.(*Client).doFollowingRedirects(0x0, 0xc208034c30, 0x480c10, 0x0, 0x0, 0x0)
	/usr/local/Cellar/go/1.4.2/libexec/src/net/http/client.go:304 +0x62
net/http.(*Client).Do(0x0, 0xc208034c30, 0x0, 0x0, 0x0)
	/usr/local/Cellar/go/1.4.2/libexec/src/net/http/client.go:174 +0xa4
github.com/Shopify/themekit.ThemeClient.query(0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, ...)
	/Users/csaunders/development/golang/src/github.com/Shopify/themekit/theme_client.go:230 +0x28b
github.com/Shopify/themekit.func·020()
	/Users/csaunders/development/golang/src/github.com/Shopify/themekit/theme_client.go:95 +0x5e
created by github.com/Shopify/themekit.ThemeClient.AssetList
	/Users/csaunders/development/golang/src/github.com/Shopify/themekit/theme_client.go:112 +0x18b

[...]

PR #98, opened 2015-09-18
Ensure default command

csaunders:
Fixes https://github.com/Shopify/themekit/issues/89

Though should it perhaps default to `--help` when no arguments were provided instead?

### DIFF
--- a/cmd/theme/main.go
+++ b/cmd/theme/main.go
@@ -21,9 +21,6 @@ const updateAvailableMessage string = `| An update for Theme Kit is available |
 |                                      |
 | theme update                         |`
 
-const commandDefault string = "download [<file> ...]"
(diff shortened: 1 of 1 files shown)


csaunders:
please review @chrisbutcher 
/cc @jtgrenz

chrisbutcher:
:ship:

jtgrenz:
no argument should output the `--help` I think.

csaunders:
I've made some updates to the default operation. If no arguments are provided the user is informed they must provide one and the `--help` text is printed to screen.

I've also removed the default command since that's no longer going to be the case when no arguments are provided.

jtgrenz:
beautiful. 

:sheep:

Issue #107, opened 2015-10-28
[Doc] Manual installation instructions PATH line


NathanPJF:
In the [manual instructions](http://themekit.cat/install/#installing-on-mac-os-amp-linux) for installing Theme Kit, it has instructions to modify your profile by adding `export PATH=$PATH:~/Applications/bin` to the bottom of the file.

The problem with this is if you have the **shopify_theme** gem already installed, the `theme` command will still be based on the gem and not theme kit.  

To avoid having to explain the theme gem or giving more details on editing your profile, this can be solved by just re-writing the `export` line to:

```
export PATH=~/Applications/bin:$PATH
```

Big ups to @jtgrenz for telling me what the gnomes in my computer were actually doing.

tanema:
fixed in 92e031dc28ae5f34594ed224fb47eba520617620

PR #108, opened 2015-10-29
make python install align with manual instructions

jtgrenz:
The manual OSX install instructions tell users to create a `/bin` folder inside `~/Applications` to install the theme executable, but the python installed adds the executable to `~/.themekit`. They should probably be the same path regardless of how its installed. This PR just edits the install path to match 

Semi-related to this issue https://github.com/Shopify/themekit/issues/107

@csaunders

### DIFF
--- a/install
+++ b/install
@@ -64,7 +64,7 @@ class Releases(object):
         return HTTPSClient.get(Releases.FEED_HOST, Releases.FEED_PATH)
 
 class Installer(object):
-    def __init__(self, ziplocation, installlocation="~/.themekit"):
+    def __init__(self, ziplocation, installlocation="~/Applications/bin"):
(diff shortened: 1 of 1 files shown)


csaunders:
I'd rather just fix the documentation to tell people to do that instead. The hidden folder is a bit of a nicer solution than this.

I'll update the docs to reflect that.

jtgrenz:
I thought I had read somewhere that Apple was recommending people place files in the ~/Applications folder rather than the home folder and I figured thats why you made the change. Can't for the life of me find any refference to this though, so maybe I dreamt it.  

You can probably close this PR then too https://github.com/Shopify/themekit/pull/109 since its related

PR #113, opened 2015-12-02
Moving types into packages

ilikeorangutans:
Moving some types into their own packages to declutter the main directory; ran golint and fixed a few minor things. 

@csaunders


csaunders:
:+1: :+1:

PR #114, opened 2015-12-02
Adds test task to makefile

csaunders:
We should have an easy way to run all the tests. With the creation of several sub-packages, we want to make it easier to add verify everything is still working.

This should also make it possible to add TravisCI support!

Please review @ilikeorangutans


ilikeorangutans:
:+1:

PR #115, opened 2015-12-02
Fixing non-ok responses

ilikeorangutans:
Fixed issue where error responses caused weird errors; moved `Theme` into `theme` package. 

@csaunders


ilikeorangutans:
I realize this fixes the issue only in a specific use case: when downloading assets. :/

csaunders:
:ok:

PR #120, opened 2015-12-11
Cleaned up directory scanning for watching, fixed minor bugs

ilikeorangutans:
Tried to find an issue with the filewatcher and ended up streamlining the filewatcher functions. They used to create a separate filewatcher for each directory and merged multiple channels to serialize all file system events into asset events. 

Instead, only a single file watcher is created and a single channel that produces filesystem events. 
Also found a potential aliasing issue, and a potential JSON binding issue. 

@csaunders


csaunders:
Nice :heart:

PR #121, opened 2015-12-11
Added OS and architecture to user-agent string

### LINKED ISSUE #101: Include Platform in UserAgent string
If the platform is encoded in the string it can assist with usage statistics.

ilikeorangutans:
Fixes #101. Not sure about the format of the string but at least it
contains arch and os now.

@csaunders


csaunders:
<3

Issue #116, opened 2015-12-04
Theme Watch - Too many files open for 249 files


sogko:
Theme watch fails when I have 249 files in the theme folder.

```
2015/12/04 03:19:01 open <..redacted_path...>/templates: too many open files
```

When  I moved out some asset files out (bringing the file count down to 218 files), theme watch works.

I tried to add the assets folder to `ignore_files`, it didn't help though, which is kind of weird. Is my config for ignore_files wrong? 

config.yml
```yml
development:
  access_token: -------
  store: ---
  ignore_files:
    - .gitignore
    - .git/
    - .idea/
    - assets/
  refill_rate: 2
  bucket_size: 40
  concurrency: 2
```

Platform: OSX 10.11.1

Cheers!

ilikeorangutans:
Can confirm, this is a repeatable bug.

ilikeorangutans:
@sogko Are you on a Unix like system (linux, mac os x)? This might be your shell's file descriptor limit. Try running `ulimit -n` which should return you the number of file descriptors that a process can use in your shell. Default for mac os x is 256 I believe which fits the described issue nicely. 

Try increasing the limit by running `ulimit -n 1024` and see if that fixes the issue.

sogko:
@ilikeorangutans Yeap, I'm on a Mac OSX; that explains a lot.
`ulimit -n 1024` did fix the issue (it was at `256` before).

Thanks! 👍🏻

PR #122, opened 2015-12-14
Automatically setting file limits on startup for unix like platforms

ilikeorangutans:
Automatically set file limits on startup for unix (non-windows) systems. See #116. 

@csaunders do we have any windows people here that could do a quick smoke test for us?

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -1,6 +1,6 @@
 source 'https://rubygems.org'
 
-ruby "2.2.2"
+ruby "2.2.3"
 
 gem 'dotenv'
 gem 'fog-aws'
--- a/Makefile
+++ b/Makefile
@@ -4,14 +4,14 @@ all:
 	for subproject in $(SUBPROJECTS); \
 	do \
 		mkdir -p build/development; \
-	  godep go build -o build/development/$${subproject} cmd/$${subproject}/*.go; \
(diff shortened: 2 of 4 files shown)


csaunders:
I believe @bman-80stees was using windows to do theme development.

csaunders:
@ilikeorangutans you can also get your hands on a Windows VM from microsoft and validate that it works the way you expect.

https://dev.windows.com/en-us/microsoft-edge/tools/vms/mac/

bman-80stees:
I use this for Shopify theme development on Windows 7. Though currently I'm using a version from way back in the spring of this year(2015), think it was in the early stages of this just after the name changed from phoenix. It still works for me without a hitch other than if I don't see a change when I refresh after I upload a file, since there wasn't error reporting that Shopify rejected the file I tried because of a liquid error in themekit.  If I update my version of themekit, I want to be sure I can continue to use it without all the different bugs I've noticed over the past several months that I get emailed about from this repo. thanks.

ilikeorangutans:
@csaunders I've tested in a windows vm, seems to work fine. 
@bman-80stees Somewhat unrelated to this, but you can get older versions of themekit from the [releases](https://github.com/Shopify/themekit/releases) page.

PR #123, opened 2015-12-14
expanded list of supported folders to include sections and blocks

stevebosworth:
@csaunders @ilikeorangutans 

Quick fix to support some new folders we need for themes going forward.  Let me know if there's any updates needed for tests and the like.

### DIFF
--- a/file_watcher.go
+++ b/file_watcher.go
@@ -14,7 +14,7 @@ import (
 
 const EventTimeoutInMs int64 = 3000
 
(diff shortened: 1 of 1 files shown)


ilikeorangutans:
Looks good to me and tests are passing for me.

cshold:
:+1:

rickydazla:
What are `sections` and `blocks`?

cshold:
Some experimental folders we're working with internally right now. We added them here so we didn't have to create a private fork of the repo.

PR #133, opened 2016-01-13
Using WriteString instead of Write([]byte)

chrisbutcher:
io.WriteString() is more optimized for cases when the argument given is already evaluated as a string. This PR uses it any place we'd otherwise do `Write([]byte(fmt.Sprintf("Foo %s", "bar"))) etc`. Cleaner and faster!

+ Lint-based style changes

@ilikeorangutans for review

### DIFF
--- a/cmd/theme/main.go
+++ b/cmd/theme/main.go
@@ -5,12 +5,13 @@ import (
 	"errors"
 	"flag"
 	"fmt"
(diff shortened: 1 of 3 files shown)


ilikeorangutans:
Nice, love the cleanups. Didn't know about `io.WriteString`. 

:+1:

PR #134, opened 2016-01-15
Cleanups

ilikeorangutans:
Various minor fixes and improvements: 

* `install` target for Makefile that installs the binary. Got tired of always doing that myself. 
* removed some debug output in leaky bucket
* moved loading assets from a local directory out of `ThemeClient` into its own method next to Asset. 

Review: @chrisbutcher


chrisbutcher:
Lookin' good!

PR #135, opened 2016-01-17
Refactoring command mapping, configuration/flags, and addressing linting warnings

chrisbutcher:
All tests passing, but we only have 47% test coverage. Let's discuss the path that these changes put us on this week @ilikeorangutans, but more so what further cleanup and idioms can be attained going forward.

The biggest changes here are:
* Consolidating to a single Args struct (which can be better understood and deconstructed into concrete types, rather than mapping strings to `interface{}`)
  * This will give us a chance to differentiate arguments to commands from the context and environment around the command(s) being run.
  * We can also revisit the topic of good defaults, and determine whether any commands need their own, specific defaults, or whether there's a single set of defaults good for all commands.
  * The args parsing stage should also be where all input validation is done, so we can abort early before any commands attempt to run if anything is amiss.
* Removing common.go functions to extract parsed fields from removed maps.
  * There may be other unused functions in the app that we can remove to reduce clutter.
* For the most part, `*Command` functions no longer take and further verify args params, before doing any real work with them.
* Adding comments, adhering to linting warnings (0 warnings now)

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -2,5 +2,4 @@ build/
 nanoc/
 .env
 tags
-
 .DS_Store
--- a/cmd/theme/main.go
+++ b/cmd/theme/main.go
@@ -24,15 +24,6 @@ const updateAvailableMessage string = `| An update for Theme Kit is available |
 
 var globalEventLog chan themekit.ThemeEvent
 
-var permittedZeroArgCommands = map[string]bool{
-	"upload":   true,
-	"download": true,
-	"replace":  true,
-	"watch":    true,
-	"version":  true,
-	"update":   true,
-}
-
 var commandDescriptionPrefix = []string{
 	"Usage: theme <operation> [<additional arguments> ...]",
 	"  Valid operations are:",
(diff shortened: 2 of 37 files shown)


cmd/theme/main.go
```diff
@@ -146,97 +146,77 @@ func main() {
...
+	result = commands.DefaultArgs()
 	currentDir, _ := os.Getwd()
 	var allEnvironments bool
-	var environment, directory, notifyFile string
-	var environments themekit.Environments
 
 	set = makeFlagSet(cmd)
-	set.StringVar(&environment, "env", themekit.DefaultEnvironment, "environment to run command")
+	set.StringVar(&result.Environment, "env", themekit.DefaultEnvironment, "environment to run command")
 	set.BoolVar(&allEnvironments, "allenvs", false, "start watchers for all environments")
```
ilikeorangutans: Is that flag actually being used?

cmd/theme/main.go
```diff
@@ -146,97 +146,77 @@ func main() {
...
+	if len(result.Environment) != 0 && allEnvironments {
+		result.Environment = ""
 	}
 
-	result["notifyFile"] = notifyFile
-	environments, err := loadEnvironments(directory)
+	environments, err := loadEnvironments(result.Directory)
 	if allEnvironments && err == nil {
-		result["environments"] = environments
+		result.Environments = environments
```
ilikeorangutans: As discussed, we should look into passing an actual instance of Environment instead of the string. I think that would be really neat! :)

chrisbutcher: Agreed!

cmd/theme/main.go
chrisbutcher: @ilikeorangutans Yes, it is: https://github.com/Shopify/themekit/pull/135/files#diff-1480839ab7d21ce62f762376dde0c0a0R181

chrisbutcher:
@stevebosworth Do you mind building from this branch, and tophatting your workflow with it tomorrow / this week? I mainly just want to see if anything blows up immediately: if it does, feel free to paste any stacktraces etc here. Thanks!

stevebosworth:
On it!  @Shopify/themes-fed if anyone else can build themekit from source, you should :tophat: as well

cmd/theme/main.go
chrisbutcher: I'll be addressing the still-remaining duplicated configuration initialization and checking logic, as well as making concrete types out of all these cli args / config.yml values after this PR.

stevebosworth:
@chrisbutcher used for the last few days and everything seems good!  :ship:

chrisbutcher:
@stevebosworth Great, thanks! Will get some additional code review and will ship a release with this and https://github.com/Shopify/themekit/pull/145 soon

.gitignore
```diff
@@ -2,5 +2,6 @@ build/
 nanoc/
 .env
 tags
+<<<<<<< HEAD
```
ilikeorangutans: Merge conflict?

cmd/theme/main.go
```diff
@@ -50,79 +41,84 @@ var permittedCommands = map[string]string{
...
-
-	for cmd, desc := range permittedCommands {
-		commandDescription[pos] = fmt.Sprintf("    %s:\n        %s", cmd, desc)
-		pos++
-	}
-
-	return strings.Join(commandDescription, "\n")
-}
+// CommandParser maps functions that take a command name and string args, to a prepared Args struct and set of CLI flags
+type CommandParser func(string, []string) (commands.Args)
```
ilikeorangutans: Just a thought... is it parsing commands? Or is it parsing arguments? I mean the return type is Args, so maybe it should be an ArgsParser?

cmd/theme/main.go
```diff
@@ -50,79 +41,84 @@ var permittedCommands = map[string]string{
...
-func setupErrorReporter() {
-	themekit.SetErrorReporter(themekit.HaltExecutionReporter{})
-}
+// Command maps command string names to Commands that return a done channel, that is closed when the command operations are complete
+type Command func(commands.Args) chan bool
 
-func setupGlobalEventLog() {
-	globalEventLog = make(chan themekit.ThemeEvent)
+// CommandDefinition ...
+type CommandDefinition struct {
```
ilikeorangutans: Love this!

.gitignore
chrisbutcher: haha nice

cmd/theme/main.go
chrisbutcher: :+1:

cmd/theme/main.go
chrisbutcher: :heart:

ilikeorangutans:
Don't have time to do a thorough review right now, but I skimmed it. It looks good, a few minor comments, but otherwise :+1: . Good work!

Issue #105, opened 2015-10-17
Invalid configuration values should fail gracefully


Leland:
This is a corollary to #89.

Running themekit with invalid values in the configuration file causes commands to spew unhelpful errors for each of the files in the request. This prevents the user from diagnosing the root cause of the error.

Case in point: I got this set up initially using a private app's API Key, instead of it's password. When ran, I got ~50 errors denoting:

> [401]Could not perform Update to {FILENAME}
> json: cannot unmarshal string into Go value of type themekit.AssetError"

I would never have figured this out unless I had stumbled across issue #104.

A helpful error dialog for each possible misconfiguration of values would greatly alleviate users troubleshooting future issues.

PR #139, opened 2016-02-02
Differentiate AssetErrors from ApiError

chrisbutcher:
Fixes https://github.com/Shopify/themekit/issues/105

This should at least result in a better error message in this scenario - let's discuss more tomorrow.

cc: @ilikeorangutans


chrisbutcher:
On second thought, we should probably just expand this section to deal with 401 responses specifically, and not treat them like 400 bad requests, or whatever typical asset errors are.
https://github.com/Shopify/themekit/blob/master/theme_event.go#L67-L69

Issue #104, opened 2015-10-13
Rename access_token to password in config.yml


cshold:
Common issue when getting setup for the first time is for the dev to use Shopify-generated API key instead of the password from their private app page. Any other possible fixes?

cc @stevebosworth @t-kelly @AWaselnuk

AWaselnuk:
Could be as simple as including a comment in config-sample.yml if you want to stay consistent with theme kit args.

csaunders:
We go back to using "password"?

The private apps page should really be updated in my opinion. For private apps the API Key isn't actually used for anything.

/cc @jamesmacaulay

stevebosworth:
I'm cool with going back to `password`.  Most are used to that from the bad old theme gem days

zachharkey:
At the very least, it would be nice to update the [docs](http://themekit.cat/docs/) to make it clear that "access_token" means Password, not API Key.

schaeken:
Ran into this issue yesterday. There is no access_token on the private app page 

![https://screenshot.click/12-46-6ip0d-104qj.jpg](https://screenshot.click/12-46-6ip0d-104qj.jpg)

Something needs to change somewhere, instead of hitting a wall and not knowing what went wrong.

stevebosworth:
Can we accept either `password` or `access_token` so this isn't a breaking change?

playmakers:
:+1:  for updating docs, just to mention `'access_token' is NOT API Key` would have helped me a lot...

chrisbutcher:
As a first step, I'm going to switch the font-family specified in the docs, since the one used (Muli) won't exist for most users, and the fallback sans-serif font don't seem to display <strong> tags properly. The docs will now look like this:
![image](https://cloud.githubusercontent.com/assets/1916444/13269504/0562030a-da56-11e5-8c7a-506cf63d3cca.png)

As for adding a comment to config-sample.yml, I think that's a good idea, too.

chrisbutcher:
Fixed in 0.3.6

PR #145, opened 2016-02-24
Support 'password' in config, interchangeable with access_token

chrisbutcher:
Part of https://github.com/Shopify/themekit/issues/104

* Adds `password` to Configure command workflow, which holds the same value sent to Shopify under the header `X-Shopify-Access-Token`.
  * The logic which allows either to be used interchangeably is [here](https://github.com/Shopify/themekit/pull/145/files#diff-bee1b19602a9f84a06f44e111f79a479R93).
  * The naming of this header is a source of confusion. In the previous themes gem, and in previous incarnations of `config.yml`, the term `password` is used. It's also used within the Shopify Private Apps UI, as seen in the related issue above.
* This PR adds support for `password` as the preferred flag name, and `config.yml` key. Eventually we should deprecate the term `access_token` themekit (a warning is added when calling `theme configure` with `-access_token=foo`).

Following my configuration refactor PR shipping, a lot of the duplication still visible in parts of this PR will be cleaned up. https://github.com/Shopify/themekit/pull/135

Please review, @ilikeorangutans cc: @stevebosworth @cshold

### DIFF
--- a/configuration.go
+++ b/configuration.go
@@ -2,16 +2,18 @@ package themekit
 
 import (
 	"fmt"
-	"gopkg.in/yaml.v1"
 	"io"
 	"net/http"
 	"os"
 	"runtime"
+
+	"gopkg.in/yaml.v1"
 )
 
 type Configuration struct {
 	ThemeId      int64    `yaml:"theme_id,omitempty"`
-	AccessToken  string   `yaml:"access_token"`
+	AccessToken  string   `yaml:"access_token,omitempty"`
+	Password     string   `yaml:"password,omitempty"`
 	Domain       string   `yaml:"store"`
 	Url          string   `yaml:"-"`
 	IgnoredFiles []string `yaml:"ignore_files,omitempty"`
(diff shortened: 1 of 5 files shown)


configuration.go
```diff
@@ -87,7 +89,14 @@ func (conf Configuration) AssetPath() string {
 }
 
 func (conf Configuration) AddHeaders(req *http.Request) {
-	req.Header.Add("X-Shopify-Access-Token", conf.AccessToken)
+	var accessToken string
+	if len(conf.Password) > 0 {
+		accessToken = conf.Password
+	} else {
+		accessToken = conf.AccessToken
```
chrisbutcher: TODO: Print deprecation warning

configuration.go
```diff
@@ -54,9 +57,12 @@ func (conf Configuration) Initialize() (Configuration, error) {
 
 	if len(conf.Domain) == 0 {
 		return conf, fmt.Errorf("missing domain")
+	} else if !strings.HasSuffix(conf.Domain, "myshopify.com") {
```
ilikeorangutans: Should we allow `.com` and `.io`?

ilikeorangutans:
One minor comment, otherwise :+1:

configuration.go
chrisbutcher: Good idea!

chrisbutcher:
Will :ship: this this week, and create a new release at that time

PR #148, opened 2016-03-01
Self documenting makefile

ilikeorangutans:
Added a help target with pretty output. All you need to do is add `## Description of target` after the target definition. 

Looks like this:
screen shot 2016-03-01 at 1 38 56 pm

@chrisbutcher


chrisbutcher:
Nice! :ship:

PR #153, opened 2016-03-30
Remove block folder from asset locations

cshold:
No need for the block folder anymore

cc @Thibaut @stevebosworth


ilikeorangutans:
:+1:

stevebosworth:
:+1:

PR #154, opened 2016-04-06
Adds UploadCommand test, fixes other tests

chrisbutcher:
cc: @ilikeorangutans 

Without approaching the issue (yet) of command functions returning `done` channels, rather than hiding channels within the implementation, or allowing callers to use channels as needed, I've added a test to UploadCommand to at least have something in place, and make small step in the right direction.

Also, all tests are now passing again. This may have changed since the original tests were written, but `assert.Equal` doesn't do a simple value comparison, but fails if an `int` (for example) is compared with an `int64`. We could do casts within the calls to `assert.Equal` to either the expected or actual values: or we could simply use `assert.EqualValues`, which [uses reflection and DeepEqual](https://github.com/stretchr/testify/blob/6fe211e493929a8aac0469b93f28b1d0688a9a3a/assert/assertions.go#L48-L64) to make sure it's comparing values of the same type.


commands/upload.go
```diff
@@ -1,28 +1,25 @@
...
 
 	done, events := args.ThemeClient.Process(files)
 	mergeEvents(args.EventLog, []chan themekit.ThemeEvent{events})
 	return done
 }
 
-func readAndPrepareFiles(filenames []string, results chan themekit.AssetEvent) {
-	for _, filename := range filenames {
-		asset, err := loadAsset(filename)
+// ReadAndPrepareFiles ... TODO
```
ilikeorangutans: What's the TODO for?

PR #155, opened 2016-04-23
Fix timeout functionality, some commands can timeout, others cannot

chrisbutcher:
cc: @ilikeorangutans to 🎩 

Aug 3: `timeout` can also be specified via config.yml


ilikeorangutans:
Gave it a quick 🎩 

- downloaded individual files
- watched directory, ensured it recognizes events
- uploaded files

👍

PR #164, opened 2016-07-21
Switching to vendor

ilikeorangutans:
Switching from godep's `_workspace` to the now standardized `vendor` dir. 

For review: @chrisbutcher


chrisbutcher:
LGTM

Issue #157, opened 2016-05-02
theme_id as string defaults to live theme


macdonaldr93:
This one was a bit of a scary one haha. We wrote a CLI tool to help us work faster. While converting JSON to YML, the `theme_id` wasn't parsed as an integer which caused ThemeKit to default to the live theme.

```yml
development:
  theme_id: "1111111111"
  password: "apipassword"
  store: "test.myshopify.com"
  ignore_files:
    - config/settings_data.json
    - config/settings.html
    - "*.png"
    - "*.jpg"
    - "*.jpeg"    
    - "*.gif"
    - "*.pdf" 
    - "*.mp4"

**Comment by jtgrenz** [comment]
for safety sake, it might be a good idea to not default to the published live theme if the id is left blank or is a non-integer. If you don't know what you are doing, or you are working on integrating another tool like above, you can replace a published theme unintentionally with no way to restore it. ThemeKit should probably parse a string for theme_id and it should probably take an explicit value for the published theme ie "published" or "live" or 0; something that is very much intentional. 

```yml
development:
  theme_id: "published"
  password: "apipassword"
  store: "test.myshopify.com"
...
```

m-ux:
@chrisbutcher @ilikeorangutans @Thibaut @davidcornu 

I think this is a pretty critical issue.  Is ThemeKit the right place to resolve this, or does this need to be handled on the OSE side?

jtgrenz:
@m-ux I'm thinking this is definitely a themeKit issue. I can't think of a way to alter this on the on the OSE without locking the published theme from API modifications which I don't think is the right  action. Working on the live theme is great in a dev shop. Although, I would LOVE the _option_ to lock a theme from any sort of updates, or updates from a  single source, thats a whole different issue. 

With this, I think a blank or non-integer theme_id should trip an error and shouldn't be caught and defaulted to the live theme. I can't think of an instance where an implied default is beneficial here where a special string ("live" or "published") wouldn't accomplish the same thing with greater safety.

jtgrenz:
scratch that, it might not be a bad idea to globally lock the live theme if the store front password is disabled. That would force all shops to either be in development mode (password up), or to use an unpublished theme for API alterations, a very good practice to encourage, if not enforce.

ThemeKit would still benefit from throwing an error though all the same, even if Shopify could prevent the published theme from being edited on a live site.

m-ux:
[...]

PR #170, opened 2016-08-09
Handle blank or non-integer theme_id specified in config.yml

chrisbutcher:
Fixes https://github.com/Shopify/themekit/issues/157

* Parse theme_id in config.yml as an integer, to ensure it's valid (at least numerical)
* Enforce required values in `theme configure` (TODO update docs)
  * domain
  * password
  * theme_id (now required)

### DIFF
--- a/commands/configure.go
+++ b/commands/configure.go
@@ -26,6 +26,12 @@ func ConfigureCommand(args Args) chan bool {
 // Configure ... TODO
 func Configure(args Args) {
 	config := args.DefaultConfigurationOptions()
+	_, err := config.Initialize()
+	if err != nil {
+		fmt.Println(err)
+		return
+	}
+
 	AddConfiguration(args.Directory, args.Environment, config)
 }
 
@@ -99,7 +105,12 @@ func loadOrInitializeEnvironment(location string) (themekit.Environments, error)
 	}
 
 	env, err := themekit.LoadEnvironments(contents)
-	if (err != nil && canProcessWithError(err)) || len(env) <= 0 {
+
+	if err != nil && !canProcessWithError(err) {
+		return env, err
+	}
+
+	if err != nil || len(env) <= 0 {
 		conf, _ := themekit.LoadConfiguration(contents)
 		env[themekit.DefaultEnvironment] = conf
 	}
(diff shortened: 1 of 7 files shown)


chrisbutcher:
Please review @ilikeorangutans 

cc: @m-ux @stevebosworth @macdonaldr93

commands/configure.go
```diff
@@ -99,13 +105,22 @@ func loadOrInitializeEnvironment(location string) (themekit.Environments, error)
 	}
 
 	env, err := themekit.LoadEnvironments(contents)
-	if (err != nil && canProcessWithError(err)) || len(env) <= 0 {
```
ilikeorangutans: Oh yeah, that's sneaky; the error just disappears.

commands/configure.go
```diff
@@ -99,13 +105,22 @@ func loadOrInitializeEnvironment(location string) (themekit.Environments, error)
...
+	if err != nil || len(env) <= 0 {
 		conf, _ := themekit.LoadConfiguration(contents)
 		env[themekit.DefaultEnvironment] = conf
 	}
 	return env, err
 }
 
 func canProcessWithError(e error) bool {
-	return strings.Contains(e.Error(), "YAML error") == false
+	if strings.Contains(e.Error(), "YAML error") == false {
```
ilikeorangutans: Hmmm, I guess the yaml parser doesn't expose custom error types? This seems brittle.

configuration.go
```diff
@@ -57,8 +58,11 @@ func (conf Configuration) Initialize() (Configuration, error) {
 	}
 
 	conf.URL = conf.AdminURL()
-	if conf.ThemeID != 0 {
-		conf.URL = fmt.Sprintf("%s/themes/%d", conf.URL, conf.ThemeID)
+
+	if themeID, err := strconv.ParseInt(conf.ThemeID, 10, 64); err == nil {
+		conf.URL = fmt.Sprintf("%s/themes/%d", conf.URL, themeID)
+	} else {
+		return conf, fmt.Errorf("missing theme_id")
```
ilikeorangutans: Here it might be useful to actually report if the theme is actually in an invalid format by appending the error message?

ilikeorangutans:
A few minor comments, then 👍

cshold:
Can we let `live` in `theme_id` upload to the live shop? That way we don't have to go and find the theme ID every time.

commands/configure.go
chrisbutcher: Agreed. We can come back to this though

configuration.go
chrisbutcher: 👍

chrisbutcher:
That should work, @cshold

Issue #179, opened 2016-08-17
Command Operations List Differs Every Time Called


chryton:
## Issue Case

The list of available operations/commands are displayed in different orders each time they are displayed.

## Examples
http://i.rocktc.de/1affA/5c2qVKlp
http://i.rocktc.de/1jvIo/4LZfhG2o

## Steps To Reproduce

1. run `theme` with no parameters; note returned operations order
1. run `theme` with no parameters again; note returned operations order
1. run `theme -h`; note returned operations order
1. run `theme -h` again; note returned operations order

## Expected Results

I'd expect that list to be consistent (alphabetical would make the most sense) between each call.

If anything, this also comes back to a documentation problem as that list is not entirely reflected in the docs: http://themekit.cat/docs/

chryton:
This was found during discussion about #166.

cc: @csaunders, @chrisbutcher

PR #191, opened 2016-09-16
Deterministic Help Message

tanema:
Fixes https://github.com/Shopify/themekit/issues/179

I thought I would start with an easy one to get familiar with the code base. This makes the help message print exactly the same every time. Thats all.


chrisbutcher:
👍

PR #192, opened 2016-09-16
Adding in theme id to the default config args

tanema:
I found that I could not use `theme configure` because the `theme_id` argument was not passed along and it was required. So I added this into the default config options so that it is passed along to the config command.


chrisbutcher:
👍

PR #197, opened 2016-09-26
Refactor source into cmd pattern.

tanema:
Generally when writing a command line tool you shouldn't have any source in the main package. This is generally done when you are writing a library. The only thing that should remain in the main package are config, docs and makefile.

I also moved bucket into the main kit package since it was only ever used with leakybucket and that was only ever used by the themekit code.

I would like to refactor more but I didn't want the PR to get too big.


tanema:
Also after this is done, I will get circle setup so we can run the tests that we actually have.

ilikeorangutans:
So you moved the bulk of the classes from main into the kit package? Did you change all the package prefixes manually or do you have a handy tool?

tanema:
Manually unfortunately. I don't know any tools that will rename things for you. That would be nice.

tanema:
Okay so I rebased so the tests work so you can see the tests are passing. I also had to change the circle build so it would work without go get so it will be testing like we run tests (with vendor) Tests all pass and the command is still entirely functional.

chrisbutcher:
LGTM

PR #199, opened 2016-09-26
Adding build flags and upx packing

tanema:
I added compiler flags and used upx to shrink our binary from 6.5M to 1.3M. This also remove stack traces that point to the machine that they were built on.


tanema:
Also as a note to @chrisbutcher deploying will take a lot longer now because upx isn't exactly fast in what it does.

chrisbutcher:
Awesome! 🚢

Issue #156, opened 2016-04-26
-allenvs fails on watch, and -env fails on upload


tyler-king:
It seems using `development` is required? We have a few sites which are clones of eachother execpt theres liquid to to detect the Shopify domain and hide/show different features. We've named the environment `ca`, `us` and `int`.

```bash
theme watch -allenvs
'' is not a valid environment. The following environments are available within config.yml:
 - ca
 - us
 - int
```

```bash
grunt release && theme upload assets/style.css.liquid assets/scripts.js.liquid -env us
...
'development' is not a valid environment. The following environments are available within config.yml:
 - ca
 - us
 - int
```

Is this a bug or `development` required?

patrick-evolabs:
I was having the same error with theme replace. After adding development it started working but theme watch -allenvs now returns : 
' ' is not a valid environment. The following environments are available within config.yml:
 - development
 - france
 - worldwide
 - taiwan

ndrewcowley:
+1 to this. If anyone knows of a fix, please share.

tyler-king:
Ive not done any Go coding but looking into the source and playing with it, it seems its just not built in right now.

It seems there needs to be a loop [here](https://github.com/Shopify/themekit/blob/master/cmd/theme/main.go#L212) to create an set of theme clients for [this function to use](https://github.com/Shopify/themekit/blob/master/commands/watch.go#L12) because [isSingleEnvironment](https://github.com/Shopify/themekit/blob/master/commands/watch.go#L32) is returning true where `args.ThemeClients` is not set, only `args.ThemeClient` (singular).

macdonaldr93:
@tyler-king for the env option, it should work if you put the environment before the list of files:

`theme upload -env=us assets/style.css.liquid assets/scripts.js.liquid`

Still a bug, but at least you can workaround until the fix.

tyler-king:
Ah ok, but what about watch? This seems to not work for watch.

Sent from my iPhone

On Aug 12, 2016, at 3:59 PM, Ryan Macdonald <notifications@github.com<mailto:notifications@github.com>> wrote:


@tyler-king<https://github.com/tyler-king> for the env option, it should work if you put the environment before the list of files:

theme upload -env=us assets/style.css.liquid assets/scripts.js.liquid

Still a bug, but at least you can workaround until the fix.

-

macdonaldr93:
`theme watch -env=staging`, for example, seems to work fine for me. There's still a bug with `-allenvs`.

[...]

PR #201, opened 2016-09-28
Fixing -allenvs flag on watch

tanema:
Fixes https://github.com/Shopify/themekit/issues/156


chrisbutcher:
🎉

Issue #177, opened 2016-08-17
Theme `upload templates/*` or `templates/**/*` Does Not Upload Customers Folder


chryton:
I've run a few releases off of the previous version without this problem. Returns with error `LoadAsset: File is a directory`.

I can upload the folder using `theme upload --env=staging templates/*/*` but that seems counterintuitive.

cc: @cshold

cshold:
@chrisbutcher @m-ux Thoughts on this? I agree, `theme upload templates/*` I would think would upload everything, not fail on folders.

chrisbutcher:
@chryton Have you tried `theme upload templates/**/*` ?
I believe this is the correct glob syntax to mean all subfolders (**) and files (*)

chryton:
@chrisbutcher Same issue: http://i.rocktc.de/15dxn/BGX4hsBW

chryton:
Weird thing is that `theme replace` _seems_ to do this correctly but that is a much slower process than upload.

chryton:
@chrisbutcher @cshold any updates on this? Still a major blocker when deploying a theme.

PR #203, opened 2016-09-28
Refactoring UpdateCommand to use the same patterns as replace

tanema:
fixes https://github.com/Shopify/themekit/issues/177

This originally started off to make it as the same pattern as replace but I found that I could really boil down the upload command to very simple code. Unfortunately I had to take out the test for it since I couldn't get around os stuff.


chrisbutcher:
🚢 💨

Issue #165, opened 2016-07-25
JSON content saved as a single line


illarionvk:
When using `./theme download` command to download all theme files, JSON file content is saved as a single line instead of pretty output. If I download it with a single file command, the JSON is saved correctly:

```
./theme download config/settings_schema.json
```

The issue occurs with settings and locale files, i.e. all files with `.json` extension. I'm running themekit 0.3.6 on a Mac.

![themekit-json-issue](https://cloud.githubusercontent.com/assets/5012724/17102165/a3d8d870-5278-11e6-80e1-0a7e4b5845d1.png)

ilikeorangutans:
Without having looked into it, could you check the line breaks in this file?

illarionvk:
`./theme download` outputs a single line JSON, no line breaks. Individual file download mode outputs indented JSON with Unix (LF) line breaks.

ilikeorangutans:
@illarionvk I just downloaded a few files (.json, .liquid) using Themekit 0.3.6 on Mac OS X and linebreaks look fine. Do you see anything I'm missing or need to do differently to reproduce this?

chryton:
@ilikeorangutans @illarionvk I was just able to recreate this via using `./theme download` to batch download the theme. When downloading via `./theme download config/settings_data.json`, it renders out with the correct line breaks. I can offer a screencast if it'd be helpful

If it matters, this was an issue with the theme gem as well.

ilikeorangutans:
Hmmm. Is it possible that the linebreaks are changed serverside?

illarionvk:
Thanks for the confirmation, @chryton.

@ilikeorangutans The issue with line breaks occurs only when you download the entire theme with `./theme download` command and only for JSON files. I guess bulk download mode and individual download mode use different functions. Could you check?

illarionvk:
I use this workaround for now:

```
./theme download && ./theme download config/*.json locales/*.json assets/*.json
```

The command downloads the entire theme first, then JSON files get downloaded using the individual download command

emiliobool:
I was trying to figure out a different issue so I set mitmproxy, and I got one line locale json files, so I removed the proxy and it was still happening, so just to make sure I wasn't going crazy I looked at older backups from a few hours ago and the files where normal, somehow the issue randomly appeared

no idea what triggers it, specially since it didn't happen just a few hours ago

illarionvk solution fixed it though

emiliobool:
[...]

PR #204, opened 2016-09-28
Pretty write JSON downloads

nicday:
Fixes https://github.com/Shopify/themekit/issues/165

Resolves by checking if the file download is a JSON asset and pretty writes. If the file is not JSON, write as normal.

@tanema @chrisbutcher

### DIFF
--- a/commands/download.go
+++ b/commands/download.go
@@ -1,7 +1,9 @@
 package commands
 
 import (
+	"bytes"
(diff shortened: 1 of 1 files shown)


tanema:
I 🎩  this. LGTM

chrisbutcher:
LGTM, thanks @nicday

Issue #205, opened 2016-09-29
Automatic Installation on Mac doesn't work


Tdum:
<img width="888" alt="capture d ecran 2016-09-29 a 16 01 34" src="https://cloud.githubusercontent.com/assets/8091423/18957098/127c027a-865e-11e6-847b-86d3395bcd8d.png">

tanema:
Hey thanks for the heads up. we are looking into this right away.

tanema:
Hey sorry about that, I have updated the website but it will take a little while before it shows so while you wait you can install with this `curl https://raw.githubusercontent.com/Shopify/themekit/master/scripts/install | python`

a-s-f:
hmm that got rid of that other error but still not working. 
![screenshot 2016-09-29 11 29 41](https://cloud.githubusercontent.com/assets/16754665/18960550/25617c56-8638-11e6-95a3-ef85a00861b6.png)

tanema:
It doesnt look like you added it to your bashrc.

a-s-f:
got it thanks @tanema  !

PR #207, opened 2016-09-29
Adding docs folder and updating install script path

tanema:
I moved all the docs to a docs folder here on master so they can be maintained at the same time as code changes. This is now supported by github.

I also updated the install script path to fix https://github.com/Shopify/themekit/issues/205


chrisbutcher:
👍 👍 👍

Issue #200, opened 2016-09-27
Long files being cut off on upload


AlfalfaAnne:
I'm having the same issue as #99  with long files (both JS and CSS).

Details:
- no errors from Terminal and the "Successfully performed Update operation for file" is returned
- long file is cut off at random (usually 4000 line mark)
- if I stop the current theme watch and rerun, it generally works, but only the first time
- only happens with extremely long files and if I shorten them, it's all good
- never had this issue with the theme gem - it's only come up since I swapped to Theme Kit

Running:
Sierra 10.12
Theme Kit v0.4.3

tanema:
Hey @AlfalfaAnne I will look into this. Since there was a previous PR that fixed this, it may just be a regression so I will look into just applying the fixes again. Cheers.

tanema:
@AlfalfaAnne do you have a file you can share with me at all to try and reproduce this? The code that fixed this issue is still in place and there should be no file limit so I will have to try and reproduce it locally.

AlfalfaAnne:
@tanema For sure - can I just send the file to timanema@gmail.com ?

tanema:
Since this is for work you can send it to tim.anema@shopify.com

tanema:
Hey @AlfalfaAnne just a quick question, what text editor are you using? I can reproduce it in vim and I see several notifications from fsnotify part way through writing the file because apparently my vim is not setup for atomic writes. So this means that the first few events are partially written files. And since we throttle these events we only get the first event with the partially written file.

I started up atom (for the first time ever) and I saved your styles.css file and it worked the first time around.

I will have to figure out how to fix this because you shouldn't have to change your env to use our tool but if you can confirm your editor that would be helpful.

AlfalfaAnne:
Hey @tanema ,

I'm using Sublime Text 3, but I'll happily try Atom for today if that helps me work faster. I'm currently working on a 1000 line JS file and I'm having to `theme upload assets/app.js.liquid` each time in order for it to not be cut off. :/ That command works every time though (unlike `theme watch`)

Not helping out my situation, I noticed my upload speed is crazy slow today.

tanema:
Yeah `upload` would always work because the file isn't partially written when it is sent to the server. I will try and come up with a solution to this soon. Thank you for your patience.

Issue #188, opened 2016-09-11
Watch removing files before uploading


Cam:
`theme watch` seems to try to remove files instead of uploading them. Pressing save twice on a file forces ThemeKit to try to remove the file then instead switch to uploading it. I'm seeing this on multiple stores/themes on El Capitan and macOS.

Cam:
#168

Issue #168, opened 2016-08-03
Theme watch requires 2 saves to update


jonnycobra:
When using `theme watch` I have to save a theme file twice. The first save removes it from the server, the second does the update. Is this a bug or an issue with my config? Using v0.3.6

jonnycobra:
Still have this issue. Normal? Or am I doing something wrong?

Cam:
#188

Cam:
I'm seeing exactly the same behaviour on v0.4.2

Issue #136, opened 2016-01-21
error message 429 - too many requests with `replace`


NathanPJF:
When running `theme replace` for themes that have many files, it's common to see a 429 error that says something like:

```
[429]Could not perform Update to templates/page.glow.liquid at *.myshopify.com
	json: cannot unmarshal string into Go value of type themekit.AssetError
```

A couple things:

1. The error message occurs when there are too many requests.  I think it would be beneficial if the error message said that instead of the "cannot unmarshal string into Go" message.

  > 429 Too Many Requests (RFC 6585)
The user has sent too many requests in a given amount of time. Intended for use with rate limiting schemes.
  > source: https://en.wikipedia.org/wiki/List_of_HTTP_status_codes

2. Would it be possible to introduce some sort of fix that either makes Theme Kit wait a set amount of time when a limit has been reached?  

3. Could there be a summary message at the end of `theme replace` and `theme update` that tells you if files were skipped and prompt you to run the command again?

### Screenshot

I have included a screenshot of a time I received an error.  In this case I was replacing a theme that has 200+ files with a theme that has ~60 files.

![http://take.ms/rRJs9](http://take.ms/rRJs9)

In this case it "timed" out after 93 actions (update, remove).  In the past we've seen it time out after ~115 actions.

NathanPJF:
May be similar to #74 

cc @ilikeorangutans @chrisbutcher

ilikeorangutans:
@NathanPJF these are actually really good suggestions! I will look into that later this week!

cshold:
I hadn't had this error in a while until this morning where it seems intermittent. Bumping it to see if we can get this looked at again

tanema:
I am seeing a lot of github traffic to this issue lately. If you are having this issue still please comment on this issue because I am(was) pretty certain that I solved this issue and I would love to know if I need to re-address it.

benVigie:
Hi @tanema , I'm experiencing this issue this morning, but it's due to my use of your tool. I'm posting the solution here in case it can help others with the same problem.

First thing first, I have `ThemeKit 0.6.12` on a Mac, several terminals open with a `$ theme watch` running all the time in one of them.

So on my side, I can **push my updates** with `$ theme upload` without any problem. But I have the `429 - too many requests` **when I want to pull theme update** with a `$ theme download`. Both actions are performed on my Shopify `development` env.  Another thing relevant it that I always getting the error when I retrieve the 42th asset each time.

[...]

PR #208, opened 2016-09-30
Throttling, Watch and Download Refactor and debouncing file watcher

tanema:
Okay this turned into a much bigger change then I expected. I had to refactor the throttler because I figured out that it was not able to properly close the done channel when it was finished the jobs it was given.

### Changed

- refactored Foreman to close the worker queue when the job queue has been closed. This enables a cascading close making commands like replace and upload complete cleanly
- refactored download to use simpler functionality
- refactored watch now to use similar patterns to upload and replace so that they should be way more understandable. There is less duplicate code and just less code in general.
- added debouncing to the file watcher to fix issue #200. The problem was that for most text editors saving large files, there would be multiple save events on the file and we would only process the first event on that file. This lead to half saved files. Now I debounced and we are processing the last write event.
- added a `NewForman()` method to the theme client as well so that we are always creating Forman objects with the configured bucket sizes.
- added logging and central event log to the theme client since this was a huge pain cascading all of the events around different channels just to log them.
- took out an incredible amount of complexity that hopefully it will make the repo a lot more nice to work with.

### Keyboard Debouncing Benifits

The debouncing of file save events also may in the future might fix the following issues as well. This is because the errors were most likely from poorly uploaded files. So it woulld throw errors on syntax, or would remove the file because some editors switch files (~filename -> filename) and it would only get the remove event.
- fixes https://github.com/Shopify/themekit/issues/200
- fixes https://github.com/Shopify/themekit/issues/168
- fixes https://github.com/Shopify/themekit/issues/188
- fixes https://github.com/Shopify/themekit/issues/97
- fixes https://github.com/Shopify/themekit/issues/79
- fixes https://github.com/Shopify/themekit/issues/209

### Throttling

Before this change, many of the commands didn't really adhere to throttling properly. This caused a lot of errors, especially the `too many requests` errors. These should be cleared up by this PR as well. 

- fixes https://github.com/Shopify/themekit/issues/136

@chrisbutcher @ilikeorangutans @jessicaxiejw


commands/replace.go
```diff
@@ -9,44 +9,27 @@ import (
...
+		}
+		for _, asset := range args.ThemeClient.LocalAssets(root) {
+			assetsActions[asset.Key] = kit.NewUploadEvent(asset)
+		}
+	} else {
+		for _, filename := range args.Filenames {
 			asset, err := theme.LoadAsset(root, filename)
-			if err == nil {
-				events <- kit.NewUploadEvent(asset)
+			if err == nil && asset.IsValid() {
```
chrisbutcher: @tanema Should we handle the `err != nil` case here as well?

commands/upload.go
```diff
@@ -9,25 +9,22 @@ import (
...
+				foreman.JobQueue <- kit.NewUploadEvent(asset)
+			}
 		}
 	} else {
-		for _, filename := range filenames {
+		for _, filename := range args.Filenames {
 			asset, err := theme.LoadAsset(root, filename)
-			if err == nil {
-				events <- kit.NewUploadEvent(asset)
+			if err == nil && asset.IsValid() {
```
chrisbutcher: Same question. Should we handle non-nil err?

kit/file_watcher.go
```diff
@@ -15,20 +13,31 @@ import (
...
-
-var assetLocations = []string{"templates/customers/", "assets/", "config/", "layout/", "snippets/", "templates/", "locales/", "sections/"}
-
-// FsAssetEvent ... TODO
-type FsAssetEvent struct {
-	asset     theme.Asset
-	eventType EventType
-}
+const (
+	debouceTimeout = 50 * time.Millisecond
```
chrisbutcher: typo 🤓

chrisbutcher:
A few minor comments, otherwise LGTM!

Can we get @m-ux @cshold @stevebosworth @macdonaldr93 to try a build off master once this merges this week?

kit/file_watcher.go
tanema: Fixed.

commands/replace.go
tanema: Fixed in both.

chrisbutcher:
🚢

Issue #175, opened 2016-08-17
theme_id validates for each environment when it doesn't need to


cshold:
With the latest 0.4.0 release, `theme_id` cannot be blank. `live` or a valid ID are the only options.

The validation seems too strong though. If I run `theme replace` I expect it to replace the default `development` shop, which it does, unless there is another environment with an invalid `theme_id`.

`theme replace` and `theme replace --env=development" fails with this config:
```
staging:
  theme_id:
  password: abcdefghijklmnop
  store: my-shop.myshopify.com

development:
  theme_id: 'live'
  password: abcdefghijklmnop
  store: my-shop.myshopify.com
```

Error message:
```
2016/08/17 16:03:03 could not load environment "staging": missing theme_id. Error: "strconv.ParseInt: parsing "": invalid syntax"
```

@chrisbutcher 
cc @Shopify/themes-fed

jeromegv:
Same issue here.

kayluhb:
I have the same issue when trying to configure a new theme following the Reame.

    $ theme configure api_key password store_domain
    missing theme_id. Error: "strconv.ParseInt: parsing "": invalid syntax"

PR #210, opened 2016-10-03
Validate Environment only when required

tanema:
fixes https://github.com/Shopify/themekit/issues/175

Just changed the validation of configurations to only happen when they are fetched using `GetConfiguration` rather than when all of the environments are unmarshaled.


tanema:
@chrisbutcher @ilikeorangutans @jessicaxiejw

chrisbutcher:
🚢

Issue #126, opened 2015-12-17
ignore_files not respected with download command


NathanPJF:
I ran `theme download --env=tpa-dec01` with the following config.yml file:

```
tpa-dec01:
  theme_id: 77036487
  access_token: #############
  store:######.myshopify.com
  ignore_files:
    - config/settings_data.json
    - "*.png"
    - "*.jpg"
```

I expected it to not download any png or jpg files, but it downloaded all of them.  Am I missing something, or is this a bug?

NathanPJF:
This also downloaded the file at `config/settings_data.json`

csaunders:
which version are you running? (`theme version`)

NathanPJF:
Theme Kit v0.3.4

NathanPJF:
Sorry - I should have specified before

Powiee:
I am also running into the same issue. "theme download config/settings.html" this command also download the file when it should have been ignored.

I tried the upload function and it is also uploading the files in the config folder. It isn't uploading the git stuff so that is good. "theme upload config/settings.html" this command is working correct and ignoring the file.

Windows 7 theme.exe
Theme Kit v0.3.5

development:
  access_token: ###
  store: ###.myshopify.com
  bucket_size: 40
  refill_rate: 2
  ignore_files:
    - .git/*
    - .gitignore
    - Config/settings.html
    - Config/*

PR #212, opened 2016-10-04
Ignored files apply to download

tanema:
Cleaned up the file watcher code a bit to use the theme clients filter and applied it to the asset lists from the server so that it filters downloads as well.

fixes https://github.com/Shopify/themekit/issues/126

cc @chrisbutcher @ilikeorangutans @jessicaxiejw


chrisbutcher:
🚢

Issue #159, opened 2016-05-04
Theme watch does not delete files


Moondrop:
Hello, the `theme watch` command does not remove files from Shopify's server if they are deleted locally. For example, if I delete an image in `/assets` it does not process. No problems with adding/uploading files however.

I had this exact same problem on the shopify_theme gem as well: https://github.com/Shopify/shopify_theme/issues/200

Cam:
Interesting. For me `theme watch` has the default behaviour of deleting any files that gets changed locally in any way at all :/

PR #213, opened 2016-10-04
Supporting remove event in watch.

tanema:
fixes https://github.com/Shopify/themekit/issues/159

I made sure that watch now removes files but also touched up the file watcher code to ensure that it will accurately respond to all events that it gets.

cc @chrisbutcher @jessicaxiejw @ilikeorangutans

### DIFF
--- a/kit/file_watcher.go
+++ b/kit/file_watcher.go
@@ -156,9 +156,12 @@ func convertFsEvents(events chan fsnotify.Event, filter EventFilter) chan AssetE
 		recordedEvents := map[string]fsnotify.Event{}
 		for {
 			currentEvent = <-events
+			if currentEvent.Op&fsnotify.Chmod != fsnotify.Chmod {
+				recordedEvents[currentEvent.Name] = currentEvent
+			}
 			select {
 			case currentEvent = <-events:
-				if currentEvent.Op == fsnotify.Chmod {
+				if currentEvent.Op&fsnotify.Chmod == fsnotify.Chmod {
 					continue
 				}
 				recordedEvents[currentEvent.Name] = currentEvent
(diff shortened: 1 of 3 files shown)


tanema:
Upon further testing I have found that the remove event doesn't successfully remove the file form the server even though it says the action completed successfully. I will look into this more.

kit/file_watcher.go
```diff
@@ -156,9 +156,12 @@ func convertFsEvents(events chan fsnotify.Event, filter EventFilter) chan AssetE
 		recordedEvents := map[string]fsnotify.Event{}
 		for {
 			currentEvent = <-events
+			if currentEvent.Op&fsnotify.Chmod != fsnotify.Chmod {
+				recordedEvents[currentEvent.Name] = currentEvent
+			}
 			select {
 			case currentEvent = <-events:
-				if currentEvent.Op == fsnotify.Chmod {
+				if currentEvent.Op&fsnotify.Chmod == fsnotify.Chmod {
```
chrisbutcher: Typo here, with the placement of `&`

chrisbutcher: Bitwise AND? :D

tanema: Yes it is a bitwise AND, because the OP can be several different ops. You need to check it this way to be acurate. Check this out: https://github.com/fsnotify/fsnotify/blob/master/fsnotify.go#L37

chrisbutcher: Actually I think it's just the spacing is off?

tanema: `go fmt` formats it this way. I have my vim setup so that is formats on save.

chrisbutcher: Ah that makes total sense. The spacing is just unusual looking

chrisbutcher:
Thanks for clarification on how go fmt formats bitwise operators, TIL.

LGTM

Issue #223, opened 2016-10-11
Multiple events in watch.


tanema:
I am seeing multiple events in watch due to low debounce timeout. this will need to be fine tuned.

PR #224, opened 2016-10-11
fixing double events and fine tuing timeout amount

tanema:
fixes https://github.com/Shopify/themekit/issues/223

I made the timeout larger and took out chmod events because they were causing double events.


tanema:

Issue #221, opened 2016-10-11
theme download does not download all files


japherwocky:
We have a theme that is also tracked in a github repository.  Something (seems like Langify or an admin) managed to delete a CSS file.

To try to check for anything else that may be missing, I cloned the repository, ran `rm assets/*`, then `theme download`.

yikes, nearly 60 files are missing!

However, I notice that some of those files are actually still being served.  I can see them in the web interface, and I can run `theme download assets/specific_file_name` to download them.

wildcard arguments, eg `theme download assets/*.png` or `theme download assets` or `theme download assets/*` don't work.  I feel like they did before this update?

this is kind of terrible guys, if we can't trust `theme download` to actually download files.

KaBrau08:
I have the same issue, and the theme watch it's failing

tanema:
What are you ignored files? Ignored files are now applicable to downloading.

japherwocky:
```
  ignore_files:
    - "*.txt"
    - "*.swp"
    - "*~"
    - ".git*"
```

a subset of files not being downloaded:

```
        deleted:    assets/gingeniousben.png
        deleted:    assets/giphy.gif
        deleted:    assets/happyending.scss
        deleted:    assets/headingVa
        deleted:    assets/heldermiranda.png
        deleted:    assets/himalayanhitman.png
        deleted:    assets/home-size-extended.jpg
        deleted:    assets/home-size-standard.jpg
        deleted:    assets/home-slider-1.jpg
        deleted:    assets/home-slider-2.jpg
        deleted:    assets/home-slider-3.jpg
        deleted:    assets/home-slider-4.jpg
        deleted:    assets/homePageChange.css
        deleted:    assets/homePageChange.css.liquid
        deleted:    assets/icheck.min.js
        deleted:    assets/icon-01.svg
        deleted:    assets/icon-03.svg
        deleted:    assets/inner-hero.jpg
        deleted:    assets/insta-2.png
        deleted:    assets/insta-3.png
        deleted:    assets/insta-4.png
        deleted:    assets/insta-5.png
        deleted:    assets/insta-6.png
        deleted:    assets/insta-7.png
        deleted:    assets/jeremyyongch.png
        deleted:    assets/jquery-migrate-1.2.1.min.js
        deleted:    assets/jquery.chic.min.js
        deleted:    assets/jquery.currencies.min.js
        deleted:    assets/jquery.cycle2.carousel.min.js
        deleted:    assets/jquery.cycle2.min.js
[...]

PR #225, opened 2016-10-12
More dependable downloads

tanema:
To make downloads more dependable I have taken out the concurrency since it doesn't really speed it up and only makes download less deterministic. I also added more error logging so I can find errors when they happen.

This may fix https://github.com/Shopify/themekit/issues/221


chrisbutcher:
LGTM #

PR #243, opened 2016-11-07
Reload watch on config change.

### LINKED ISSUE #236: Handle new config.yml on change detection (either error, warn or use new data)
I am working with several git branches, each branch with a different config.yml. In deed, each branch is a different shopify shop. With this workflow is a **really common mistake** to change branches without stopping `theme watch`. So, accidentally chaging branches will cause lot of unintended updates on my sites.

I think it is reasonable to stop themekit execution if current `config.yml` has changed.

Another workarounds would be to raise a warning (but it is not a solution for me), to prompt asking if the user want to continue with the new configuration or to operate using new configuration.

tanema:
fixes #236 

There are instances where users are checking out different branches for different shops but then the watch command would make the changes to the wrong shop.

@chrisbutcher

### DIFF
--- a/cmd/theme.go
+++ b/cmd/theme.go
@@ -116,6 +116,7 @@ func generateThemeClients() ([]kit.ThemeClient, error) {
 
 	setFlagConfig()
 
+	configPath, _ = filepath.Abs(configPath)
 	environments, err := kit.LoadEnvironments(configPath)
 	if err != nil {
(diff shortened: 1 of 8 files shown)


tanema:
More work needs to be done to this I believe because if the branch is checked out it will fire events for all files and this does not current catch that before all those changes take effect

Issue #233, opened 2016-10-26
[403]Insufficient permissions to perform Remove to layout/theme.liquid


plisvb:
Getting this error in terminal when I save a file for updating.  If I save again, it works fine.  Any idea why?

tanema:
What editor are you using and is it a large file? Some editors create a swap file of the changes, delete the current file then put the swap file in place. I am debouncing all the events but it seems like the remove is the most recent event it catches rather than the write. I will play will the debounce time for this.

plisvb:
Hi Tanema,

Using coda on the mac. it's my theme.liquid file which is 16kbs but it has issues with all files.  I save the file, it removes, I save the file again, it works fine.  Assets images etc... work perfectly

PR #245, opened 2016-11-08
Changing debounce from single timeout to a timeout for each event.

tanema:
fixes https://github.com/Shopify/themekit/issues/233

There were still problems with the file updating debouncing. I had more of an event loop pattern for my debouncing where it would check all events every second. Now I have moved toward a per event timeout which means each event will be handled one second after the last event on a file was seen. The last seen event will be handled. If this is more of a problem in the future we can provide configuration for the debounce value.

@chrisbutcher

### DIFF
--- a/kit/file_watcher.go
+++ b/kit/file_watcher.go
@@ -12,7 +12,7 @@ import (
 )
 
 const (
-	debounceTimeout = 1 * time.Second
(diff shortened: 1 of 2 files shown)


chrisbutcher:
👍

PR #254, opened 2016-11-13
Make install script work with Python 3

csears:
Current install script uses print statements without parentheses, which causes errors in Python 3...

```
$ python3 install
  File "install", line 75
    print "Downloading themekit from %s/%s" % (self.host, self.path)
                                          ^
SyntaxError: invalid syntax
```

For this new version of install, I used 2to3 to fix the print statements and a few other minor issues. It works with 2.7, 3.4 and probably any Python version in between. 

This does NOT work with 3.5 because of some breaking module name changes with httplib and urlparse.

### DIFF
--- a/scripts/install
+++ b/scripts/install
@@ -57,8 +57,8 @@ class Releases(object):
             return
         doc = ET.XML(self.__getfeed())
         entries = doc.findall(ns+"entry")
-        versions = map(lambda e: os.path.basename(e.find(ns+"link").get('href')), entries)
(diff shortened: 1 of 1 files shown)


tanema:
Thanks for your work @csears. I don't often use python so I don't know much about it but will this still work with python 2?

csears:
Hi @tanema!

Yes, it will still work with Python 2. I mentioned testing it against Python 2.7, which is the most popular version of Python 2.  Python 2.7 has been the default version on Macs since OSX 10.7.

tanema:
Hey, sorry I should have read that better. This LGTM however I will give it a try on monday before I merge it. Thanks again for your help!

PR #272, opened 2016-11-28
Installation work

tanema:
Two things accomplished with this PR,

- install script is now put into /usr/local/bin so that the user no longer needs to edit their bash settings as well as being more in line with best practices
- home brew installation formula. 

@chrisbutcher


tanema:
Right now this won't really work on linux because of permissions looking for a better solution.

tanema:
Okay I have changed how the install script is run and the permissions on the script so that is can be installed on any machine at `/usr/local/bin` however is requires users to trust us to run `sudo python`

PR #280, opened 2016-12-01
Logging work

### LINKED ISSUE #273: Put timestamp on lines
Please add the ability to display timestamps on every line of status output. This way, when working in a single screen setup and on the same file over and over I won't be wondering "Did that change go up to the server or these logs are from my old changes".

tanema:
fixes #273 

The log now looks like this
image

I also took out all the extra logging functions so that it is clear where messages are logged, either sdtout or stderr.

@chrisbutcher

### DIFF
--- a/cmd/download.go
+++ b/cmd/download.go
@@ -90,7 +90,7 @@ func writeToDisk(directory string, asset kit.Asset) error {
 		return err
 	}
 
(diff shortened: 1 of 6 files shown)


chrisbutcher:
🎉

PR #301, opened 2017-01-03
Restrict all operations to project files and non filtered files only.

### LINKED ISSUE #300: Ability to whitelist files/folders instead of ignoring
Working on a project which does not build into a `dist/` before using themekit to upload and I'm trying to whitelist the relevant folders because we have so many folders/files at the root of the project that aren't part of the theme itself.

We tried using a regex but negative lookaheads are not supported by go's regex implementation.  We also tried using negated globs to whitelist the `assets/ config/ layout/ locales/ snippets/ templates/` folders.  This works with `theme watch` but not with `theme replace`.

Our folder structure looks like this:

```
.DS_Store
.bundle/
.dev/
.git/
.gitignore
.jscsrc
.jshintrc
.ruby-version
.scss-lint.yml
.themekit-ignore
Gemfile
Gemfile.lock
Procfile
README.md
Rakefile
_sprockets/
application.yml.dist
application.yml.ejson
assets/
browserslist
circle.yml
config/
config-dev.yml
config-staging.yml
config-staging.yml.example
config.yml
dev.yml
layout/
lib/
locales/
node_modules/
package.json
service.yml
shipit.yml
snippets/
spec/
templates/
theme-builds/
```

we only want to upload:

```
assets/ 
config/ 
layout/ 
locales/ 
snippets/ 
templates/
```

tanema:
fixes #300 

Filtering out files that are not in the project workspace for all actions. This will minimize all the 404 responses from file not being in the project workspace. It will also minimize requests to Shopify which will make upload and replace faster since there are no wasted connections.

@chrisbutcher @stevebosworth

### DIFF
--- a/kit/file_filter.go
+++ b/kit/file_filter.go
@@ -100,20 +100,22 @@ func (e fileFilter) filterAssets(assets []Asset) []Asset {
 	return filteredAssets
 }
 
-func (e fileFilter) matchesFilter(event string) bool {
-	if len(event) == 0 {
-		return false
+func (e fileFilter) matchesFilter(filename string) bool {
+	if len(filename) == 0 || !assetInProject(filename) {
+		return true
 	}
+
 	for _, regexp := range e.filters {
-		if regexp.MatchString(event) {
+		if regexp.MatchString(filename) {
 			return true
 		}
 	}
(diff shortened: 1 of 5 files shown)


stevebosworth:
🎉

PR #315, opened 2017-01-16
Update docs to correct directory flag

frehner:
`--directory` is not a valid flag, however `--dir` is. Updated docs to match.


tanema:
Thanks for the work!

PR #320, opened 2017-01-19
Took out values and attachments from asset list queries

### LINKED ISSUE #318: "Client.Timeout exceeded while awaiting headers" error on OS X. Cannot download theme.
23:25:13 Assets Perform Retrieve at host <[REDACTED].myshopify.com>
	    Status: 0
	    Errors: Get https://[REDACTED].myshopify.com/admin/themes/[REDACTED]/assets.json?fields=key,attachment,value: net/http: request canceled (Client.Timeout exceeded while awaiting headers)

tanema:
fixes #318 

ThemeKit was using an API endpoint incorrectly so asset listing had to be changed to take out the attachment and value data from the list queries. (https://github.com/Shopify/shopify/issues/98379)

This will make download and replace a lot more dependable for larger themes however it does *feel* slower to download. It will however make replace much faster.

@chrisbutcher 

cc @NathanPJF @DrewMartin @nicday @Thibaut Just to let you know that I have managed a fix for this issue.

### DIFF
--- a/cmd/download_test.go
+++ b/cmd/download_test.go
@@ -41,7 +41,7 @@ func (suite *DownloadTestSuite) TestDownloadWithFileNames() {
 
 	client.Config.ReadOnly = true
 	err = download(client, []string{"output/nope.txt"})
-	assert.NotNil(suite.T(), err)
+	assert.Nil(suite.T(), err)
 }
 
 func (suite *DownloadTestSuite) TestDownloadAll() {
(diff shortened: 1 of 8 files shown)


cmd/download_test.go
```diff
@@ -30,7 +27,8 @@ func (suite *DownloadTestSuite) TestDownloadWithFileNames() {
 	defer os.Remove("../fixtures/project/assets/hello.txt")
 	defer os.Remove("../fixtures/project/assets/hello.txt")
 	client, server := newClientAndTestServer(func(w http.ResponseWriter, r *http.Request) {
-		if "fields=key,attachment,value&asset[key]=assets/hello.txt" == r.URL.RawQuery {
+		println(r.URL.RawQuery)
```
tanema: 🔥

kit/http_client.go
```diff
@@ -84,12 +84,14 @@ func (client *httpClient) ThemePath(themeID int64) string {
 }
 
 func (client *httpClient) AssetQuery(event EventType, query map[string]string) (*ShopifyResponse, Error) {
-	path := fmt.Sprintf("%s?fields=key,attachment,value", client.AssetPath())
```
tanema: This is the main line where the fields are taken out for the list query

nicday:
👍

PR #324, opened 2017-01-30
Fix typo in theme.go

petetnt:
Just something I ran into while `grep`ping the `theme --help` output for `environment` 😹


tanema:
Thanks!

PR #325, opened 2017-02-02
Verbose flag & Progress Bar

tanema:
The aim of this PR is to simplify output for download, replace, and upload. It also has the nice addition of being able to show progress. I am going to leave this PR up for a while because I am not sure if anyone wants this. I just did it while on the train to summit.

*note:* This does not change the output of watch.

**Without verbose**
image

**With verbose flag**
image

@NathanPJF  would this conflict with slate's usage?

### DIFF
--- a/cmd/bootstrap.go
+++ b/cmd/bootstrap.go
@@ -60,12 +60,14 @@ func bootstrap() error {
 		return err
 	}
 
-	kit.Printf(
-		"Successfully created theme '%s' with id of %s on shop %s",
-		kit.BlueText(theme.Name),
-		kit.BlueText(theme.ID),
-		kit.YellowText(client.Config.Domain),
-	)
+	if verbose {
+		kit.Printf(
+			"Successfully created theme '%s' with id of %s on shop %s",
+			kit.BlueText(theme.Name),
+			kit.BlueText(theme.ID),
+			kit.YellowText(client.Config.Domain),
+		)
+	}
 
 	return download(client, []string{})
 }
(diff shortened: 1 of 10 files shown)


cmd/bootstrap.go
```diff
@@ -43,11 +43,13 @@ func bootstrap() error {
 	}
 
 	themeName := getThemeName()
-	kit.Printf(
-		"Attempting to create theme %s from %s",
-		kit.YellowText(themeName),
-		kit.YellowText(zipLocation),
-	)
+	if verbose {
+		kit.Printf(
```
chrisbutcher: Maybe kit.Printf can no-op if verbose is false?

tanema: I don't really want to do that because it there might be instances where I want it to output a message even when verbose is set to false. The other way would be to have logging levels but I don't want to overcomplicate things. If it starts to get too out of hand then I will approach it again.

chrisbutcher: 👍

NathanPJF:
@tanema This should be good to merge.  For Slate we specify which version of Theme Kit we want [node-themekit](https://github.com/Shopify/node-themekit) to use - so we can always test after merge.  Or am I missing something there about node-themekit @macdonaldr93?

> I am going to leave this PR up for a while because I am not sure if anyone wants this.

I'm into it 😄 .  Right now when I run a deploy I instinctively want to walk away for coffee because I have no idea how long it will take.  Also keeps my terminal from getting cluttered 👍

macdonaldr93:
@NathanPJF not missing anything. Until we manually update node-themekit and the dependency in slate-tools, this won't be included.

PR #328, opened 2017-02-03
python 3 fix

tanema:
Fix for https://github.com/Shopify/themekit/commit/9ae31714bc1a96af2f5703049538a45f989b983d#commitcomment-20733658

"0755 is not a valid numerical literal in python 3.5. If you change it to 0o755, it will be compatible with both 2.7 and 3.5. See: https://www.python.org/dev/peps/pep-3127"


iillyyaa:
Thanks. Ship it! 👍

PR #347, opened 2017-02-22
Adding windows color support for logging to the console

tanema:
Currently color support causes output to the windows command line to seem a bit jumbled because cmd.exe doesn't know how to interpret the color codes. This fixes that so that windows developers will see the color and have properly formatted output.


tanema:
Also crazy thing I just noticed is that Matz made the colorable library.

PR #351, opened 2017-02-24
WIP - Add "customize" command

chryton:
- Work In Progress PR
- New to Go so I'm probably missing something
- Thinking about adding an option for what the URL in the editor should be (like if I want to look at a product's vs. a page's section admin)

### DIFF
--- a/cmd/customize.go
+++ b/cmd/customize.go
@@ -1,0 +1,37 @@
+package cmd
+
+import (
+	"fmt"
+	"sync"
+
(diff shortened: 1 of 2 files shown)


tanema:
I would add this functionality to the open command instead so that you can call `theme open editor` and `theme open preview` the default would be preview.

chryton:
That makes sense; I'll refactor.

chryton:
Closing in favor of #358

PR #363, opened 2017-03-08
Refactoring watch for performance

### LINKED ISSUE #349: macOS high CPU usage on `theme watch`
There is a high CPU usage when the theme watch command is left active for large periods of time. It makes my macbook fans run on high. The current solution is to exit it, and start the theme watch command again.

I know it only happens when the theme watch is active for a large period of time, but I thought I would mention it in case there is a process loop that is multiplying itself.

(I originally posted the issue [here](https://github.com/Shopify/shopify_theme/issues/235) forgetting the app has moved to this repo)

tanema:
fixes #349 

After I refactored the watching event loop so that I could more accurately profile the cpu usage actually went down to peak 7%. This could be because I have moved a large amount of variables from anonymous functions into the watcher struct. This could the GC less like to take longer in figuring out if the variables are needed or not. This is only a guess because I am still not quite able to figure out what was causing the performance issues.

### DIFF
--- a/kit/event_map.go
+++ b/kit/event_map.go
@@ -1,0 +1,50 @@
+package kit
+
+import (
+	"sync"
+
+	"github.com/fsnotify/fsnotify"
+)
+
+type eventMap struct {
+	sync.RWMutex
+	events map[string]chan fsnotify.Event
+}
+
+func newEventMap() *eventMap {
+	return &eventMap{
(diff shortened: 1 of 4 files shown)


tanema:
I refactored the notify file so that the watch event loop no longer keeps ticking for the notify file. It will only create a go routine when an event happened to touch the notify file.

With this change I have seen peak CPU at **0.5%**

Issue #369, opened 2017-03-16
Watching file changed inside a docker container


JasonShin:
Hi, I've set up a simple docker container for my theme development. Unfortunately ```theme watch``` wouldn't pick up file changes made from host machine. Is there any way to resolve this?

tanema:
Sorry you will have to give me a whole bunch more information. Why did you setup a container for development. What is the setup? Do you have your files in the container and running themekit from host? If you are running themekit from the container how are you installing themekit in your container?

tanema:
Any update?

JasonShin:
Hi, sorry about late reply. I've lost my Dockerfile but it was quite simple. It was like
```
FROM ubuntu:latest

# aptget update, essential packages
# install node npm
# install linux version of themekit

```
And I've created docker-compose that builds this dockerfile and shares volume ``` # volume share . /home ```. And I've ran the docker image by giving a interactive mode flag ``` docker run -it themekit bash```

To be honest, I've tried out this setup as an experiment.

JasonShin:
I haven't written in Go before but I've noticed it's using 'fsnotify' for watching file system change. I found this issue https://github.com/fsnotify/fsnotify/issues/152 but I guess that's not very helpful =(..  I wonder how Nodemon can achieve file watch across NFS using --legacy flag

tanema:
That is a good find. I am going to close this since it seems like there is a limitation of the file watcher and really there is no reason to be doing theme development in a container. Your theme is deployed on shopify and you don't need any other services like databases/memstores.

JasonShin:
agreed =)..

krumware:
We are new to themekit, but use containers for most things. Here it would be useful, since we can use the container to encapsulate the themekit, and allow devs to work with the themes without knowing how to install/manage themekit.  (which does sound a bit silly, since they need to learn liquid anyways, but its one less thing to think about)

In legacy mode, nodemon uses polling to detect file changes.  This is also useful in some cases where there are a large number of files to watch.
I don't have much experience with golang, but this might serve as a useful reference: https://github.com/fsnotify/fsnotify/issues/9

tanema:
@krumware I am still working a better installer for windows, would that alleviate this issue?
 Supporting a polling fallback just for container development adds significant complexity to themekit and supports a very small amount of user base.

krumware:
I think that it would. Even a powershell script or using a package manager like Chocolatey would help.

[...]

PR #390, opened 2017-05-03
Powershell windows install

tanema:
rel: https://github.com/Shopify/themekit/issues/369

I added a `ps1` script to automatically install themekit on windows. There was a problem with the msi where it would truncate the PATH variable. This should make that no longer a problem. This also pulls the most recent version so doesn't have to be changed for every release.

These would be the new directions on the docs website. 
image

@krumware would this be a helpful solution for you?


JasonShin:
we should register this on Chocolatey packages =)

Issue #392, opened 2017-05-11
`theme watch` doesn't delete a file in store admin


alvin-ahn:
Hi Team!

If I delete a file from local while `theme watch` is on, it will empty the file in a store admin but the file itself is not getting deleted. 

Thank you!

tanema:
Hey @alvin-ahn can you please tell me a few bits more of information?

- OS
- Shell
- Themekit version
- Text editor
- How did you delete the file (through your editor/command line/file exporer)

cheers,

tanema:
I cannot reproduce this using command line remove using Finder, command line and/or vim. Can you please give me any extra information?

alvin-ahn:
@tanema 

Hi Tim!

Sorry it was my first time creating a GH issue for this repo so didn't know the format.

OS: macOS Sierra Version 10.12.4
Shell: zsh
Themekit version: 0.6.11
Text Editor: Both Atom and Sublime
I deleted the file through my editor.

Deleting a file using Sublime Text: https://screenshot.click/16-33-iv47m-r2ozo.mp4

The file still exist but empty https://screenshot.click/16-34-0n44t-idlrg.mp4

tanema:
Ah okay I get it now. It turns out that Atom/Sublime Rename the folder to go in the trash rather than deleting them. I will have to figure out a cross platform way to catch when this happens

tanema:
This has actually surfaced the problem that renaming doesn't work either.

PR #395, opened 2017-05-16
Handling renaming events better

tanema:
fixes https://github.com/Shopify/themekit/issues/392

**Problem**
In Atom and Sublime Text when deleting a file instead of a delete event, a rename happens. The editors rename the file to be in the trash can. Also while testing this it seems that the same would happen when renaming a file. The renamed file would have no content with the original filename and stay existing with the newly named file

**Solution**
*Change rename events to delete files.*
This solves renaming because it will delete the first file then create a new file with the new name and content.

**Problems**
This may cause confusion if a new developer tried deleting or renaming critical files because it will fire a delete action on the critical file and fail. Then it will succeed in creating the new file. Unfortunately I cannot see a way around this.

This has been validated for:

- [x] renaming
- [x] deleting in sublime
- [x] deleting in atom

cc @NathanPJF

### DIFF
--- a/kit/file_watcher.go
+++ b/kit/file_watcher.go
@@ -203,7 +203,7 @@ func (watcher *FileWatcher) handleEvent(event fsnotify.Event) {
 		return
 	}
 	eventType := Update
-	if event.Op&fsnotify.Remove == fsnotify.Remove {
+	if event.Op&fsnotify.Remove == fsnotify.Remove || event.Op&fsnotify.Rename == fsnotify.Rename {
 		eventType = Remove
 	}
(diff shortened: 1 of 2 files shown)


tanema:
An example of a rename event 
image

NathanPJF:
> This solves renaming because it will delete the first file then create a new file with the new name and content.

Love it!

/cc @Shopify/themes-fed

alvin-ahn:
Thank you @tanema !!
You are amazing! 👍

tanema:
Hey @alvin-ahn, the fix should probably be release in the next week or two.

Issue #388, opened 2017-04-26
Theme Upload locales error


overturex:
Hi There,

I am getting the following error when using Theme Upload.

Asset Perform Update to locales/zh_CN.json at host staging-ca-sengled-com.myshopify.com
        Status: 422 Unprocessable Entity
        Errors: Invalid locale name
                Key "Commands:" contains invalid characters
                Key "Options:" contains invalid characters
                Key "Examples:" contains invalid characters
                Key "default:" contains invalid characters
                Key "choices:" contains invalid characters
                Key "Not enough non-option arguments: got %s, need at least %s" contains invalid characters
                Key "Too many non-option arguments: got %s, maximum of %s" contains invalid characters
                Key "Missing argument value: %s" contains invalid characters
                Key "Missing required argument: %s" contains invalid characters
                Key "Unknown argument: %s" contains invalid characters
                Key "Invalid values:" contains invalid characters
                Key "Argument: %s, Given: %s, Choices: %s" contains invalid characters
                Key "Argument check failed: %s" contains invalid characters
                Key "Implications failed:" contains invalid characters
                Key "Not enough arguments following: %s" contains invalid characters
                Key "Invalid JSON config file: %s" contains invalid characters
                Key "Path to JSON config file" contains invalid characters
                Key "Show help" contains invalid characters

I don't even have `zn_CN.json` in theme's locales folder. Same error is also given for bunch of other locales files such as `locales/be.json`, `locales/zh.json`, `locales/tr.json` etc.

tanema:
Hey @overturex I have a few questions to figure this out. Does `locales/zh_CN.json` exist on your remote theme (the online editor)? Do any of those labels seem familiar? Does everything else success when uploading?

Can you also please email me at tim.anema@shopify.com your store domain and which theme you are developing with so I can test this out?

tanema:
Hey @overturex I contacted support to find your shop. I was unable to reproduce this bug. Are you sure that those file do not exist in your local development files. There seems to be no other reason why you would get this error. Theme kit recursively searches your files for items to upload. Do you still get the same error when you run `theme upload locales/*`

tanema:
Hey @overturex, I am just checking back in with this issue. If you can, please get in touch with me so I can resolve this issue.

overturex:
[...]

PR #407, opened 2017-06-09
Restricting the path checking to be more strict about project paths only

tanema:
fixes https://github.com/Shopify/themekit/issues/388

It was noticed that some paths would match up with paths in the node_modules folder causing weird errors while uploading. This exposed the fact that path matching was too fuzy and probably eating up extra cpu just checking large node modules structures. 

This PR restricts all path matching, any function that makes sure a file is valid for upload is restricted to the root directory as specified by the users configuration. It was also exposed that file filtering probably didn't work through symlinks either so that was fixed in this PR as well.

This is still WIP because it is quite a substantial change and could have some potential for bugs so I will need to test it thoroughly. The tests were not changed much and they all pass however I would like to do trial runs by hand.

### DIFF
--- a/cmd/bootstrap_test.go
+++ b/cmd/bootstrap_test.go
@@ -39,7 +39,7 @@ func (suite *BootstrapTestSuite) TestBootstrap() {
 	err := bootstrap()
 	assert.NotNil(suite.T(), err)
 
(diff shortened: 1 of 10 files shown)


tanema:
I have tested this by hand for all actions and I believe that this is ready.

PR #420, opened 2017-06-27
Remote file checking

### LINKED ISSUE #330: Detecting remote changes to theme
One of the biggest pain-points when I work with a Shopify client is:

I have my own deployment process using the CLI, but clients want the flexibility to make changes in the Shopify admin as well. When they make changes and forget to tell me, their changes get overwritten when I deploy. The other way is to download the whole theme and merge the changes into my version control system.

Could we get a feature where we get notified of remote changes to the theme? Maybe check the timestamp of the last remote update compared to a timestamp of last upload kept locally? Ideally we'd be able to know which files were updated remotely. 

Thanks for all your work!

tanema:
fixes #330

This is a WIP PR to show that there is some progress on this issue. It has now reached the proof of concept stage. There are still some major decisions to be made about this 

Design/Planning notes:

- [x] `download` fetch versions from all environments on if the version is not defined.
  - [x] optimize by only downloading files that have changed.
  - [x] prune lock file so it doesn't grow endlessly.
- [x] `upload`/`replace`/`remove`, preflight checks can be done so that they don't fail in partial states
  - [x] add `force` flag to override
  - [x] preflight finds a file that we don't have on our machine: warn about new file
  - [x] preflight warns if we have a version but the shopify doesn't have that file. File is deleted
  - [x] preflight outputs a list of files that have been changed, added or deleted since last check.
  - [x] output recommended ways of merging these changes.
  - [x] update the version when uploaded and replaced of the changed file. 
  - [x] If the version does not exist for the other environments fetch it and persist to ensure versions even in environments where they previously did not have this functionality. This will make sure that if files are changes in development then those changed file will still be checked before deploy.
  - [x] optimization: `upload`/`replace` would only have to push/delete files that have changed
- [x] `watch` throw error and prompt the user to download the file
  - [x] add a `force` flag to watch to supress checking.
- [x] this can all be stored in a `theme.lock` file in the same directory as the `config.yml`
- [x] ensure errors adhere to stderr/stdout to maintain slate output.
- [x] tests will have to be re-written extensively
- [x] Move all output out of `kit` and only in `cmd` so that logging can be tested
- [x] Add comment to `theme.lock` file that it is generated and should not be edited

#### Testings issues:

- [x] `watch` not outputting errors, making it seems like the command was hanging
- [x] `download` doesn't seem to update files. (I believe this was a result of mishandled go routines)
- [x] `replace` throws error if you do not have a `settings_data.json`
- [x]  cannot ignore `settings_data.json` on `upload`/`replace`
- [x]  `Remote files are inconsistent with manifest` is an unclear error message.
- [x] manifest warning from `replace` even when there is no manifest
- [x] manifest not generated on `replace` (if no action has happened, should we generate the whole manifest?)
- [x] Remove environments from manifest that are no longer in config
- [X] some environments pruned without needing to be
- [X] all environments need to be valid before the commands can run.

### Other PRs before new version release
- [x] Add documentation for theme.lock on another branch to be merged after this PR has shipped.(https://github.com/Shopify/themekit/pull/426)
- [x] Move kit testing over to use new `kittest` package so that tests are uniform(https://github.com/Shopify/themekit/pull/423)

### Potential Pain points
- How flags and environment variables kind of change with this PR. Since this work will load every environment to check version, the flags and env vars will get applied to all the environments making some values not make sense. For instance it would not make sense for a password to be applied across all environments. Flags and env vars may have to be applied to active environments only.

Resources:

Watch error:
screen shot 2017-06-27 at 1 05 25 pm
Serialized themekit data format:
image
Preflight version check output:
image


blboyle:
Hey rest of @Shopify/msfed check out what Tim's up to.

PR #450, opened 2017-08-23
Skipping invalid environments in the manifest, allowing reusable config

### LINKED ISSUE #446: environment validity check flags repeatable yaml nodes as invalid
This suddenly started happening when I updated from `v0.6.12` to `v0.7.2`. No changes to my config were made.
### What I expected to happen
Specified files upload to my theme.
### What actually happened
`[config] Invalid configuration: missing theme_id`
### Steps to reproduce

1. Define an environment with a `theme_id` in `config.yml`
2. Run the `upload` command with the `--env` or `-e` flag pointing to that environment.

### My Environment

OS (including version): MacOS Sierra 10.12.6 (16G29)
Themekit version: `v0.7.2`
Ignore Files(if any): 
```
ignore_files:
    - src/
    - node_modules/
    - Gulpfile.js
    - package.json
```

tanema:
fixes #446

Currently themekit checks all environments strictly for version checking however this does not allow for reusable config like this:

```
production: # LIVE WEBSITE
  theme_id: ********
  readonly: true # safety switch; toggle to deploy
  << : *common
config: &common
  password: *********************
  store: storename.myshopify.com
  ignore_files:
    - src/
    - node_modules/
    - Gulpfile.js
    - package.json
```

Because the `config` environment would be invalid. I don't love this because if some of the environments are invalid I won't be able to backfill the manifest with versions. However when upload/replacing on different environments, the backfills aren't *really* necessary. I was just being greedy with that information.

### DIFF
--- a/cmd/command_arbiter.go
+++ b/cmd/command_arbiter.go
@@ -69,11 +69,7 @@ func (arbiter *commandArbiter) generateThemeClients(cmd *cobra.Command, args []s
 		isActive := arbiter.shouldUseEnvironment(env)
 		config, err := configEnvs.GetConfiguration(env, isActive)
 		if err != nil {
-			return fmt.Errorf(
-				`[%s] %s All environments are required to be valid so that asset versions can be validated`,
-				green(config.Environment),
-				yellow(err.Error()),
-			)
+			continue
 		}
 		if arbiter.disableIgnore {
 			config.IgnoredFiles = []string{}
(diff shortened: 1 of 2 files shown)


neycodes:
@tanema I'm still getting the same error on v0.7.2.

```
Invalid configuration: missing theme_id,missing store domain,missing password All environments are required to be valid so that asset
 versions can be validated
```

tanema:
Yes because this change was made after 7.2 and it will be in the next release.

neycodes:
@tanema Ah, right. Thank you!

PR #475, opened 2017-10-06
Typo fix at cmd/upload.go

jbruni:
### WHAT are you trying to accomplish with this PR?

It is too complex for me to explain... sorry.

### WHY are you choosing this approach?

Years of advanced deep research led me to believe this is the most appropriate approach.

### HOW is this accomplished?

You wouldn't understand if I tried to explain.

### Anything that could go wrong?

Of course not!

### Checklist

- [X] It is safe to simply rollback this change.
- [X] I have :tophat:'d these changes by using the commands I changed by hand


tanema:
😓  I am so glad you took care of this change, I never would have figured that out myself. Thanks! :D

PR #486, opened 2017-11-03
Adding desktop notifications to the watch command

darol100:
### Adding desktop notifications to the watch command
I was running a `theme watch` and I had an error in my liquid template and I did not notice. However, this error was caught by the logs in the watch.  So this pull request is going provide a desktop notification when the watch command starts and also if there is an error in any liquid template.

I use [Laravel Mix](https://github.com/JeffreyWay/laravel-mix) for compiling my CSS/JS using webpack and they have desktop notifications whenever the watch start and whenever there is an error and I found it very helpful.

Here is screenshot how it looks right now:

screen shot 2017-11-03 at 3 13 31 pm

screen shot 2017-11-03 at 3 28 40 pm

### Checklist

- [x] It is safe to simply rollback this change (e.g. changes to `theme.lock` that might corrupt a users workflow).
- [x] I have :tophat:'d these changes by using the commands I changed by hand
- [ ] I have run successfully run linting and tests on the codebase. **I could not run make check, it gave me some error**
- [ ] I have added tests for the changes I have made.  **I do not think is need it**
- [x] I have added a dependency to the project

### Things that can improve this PR
- We could change the icon to be Shopify icon
- We could consider adding a desktop notification whenever anything gets upload it.

### DIFF
--- a/cmd/watch.go
+++ b/cmd/watch.go
@@ -6,16 +6,21 @@ import (
 	"log"
 	"os"
 	"os/signal"
-
 	"github.com/spf13/cobra"
-
+	"github.com/0xAX/notificator"
 	"github.com/Shopify/themekit/kit"
+	"path/filepath"
+	"runtime"
 )
 
 var (
 	signalChan   = make(chan os.Signal)
 	reloadSignal = make(chan bool)
 	errReload    = errors.New("Reload Watcher")
+	notify       *notificator.Notificator
+	_, b, _, _ = runtime.Caller(0)
+	basepath   = filepath.Dir(b)
+
 )
 
 var watchCmd = &cobra.Command{
@@ -64,6 +69,8 @@ func watch() error {
 			continue
 		}
 
+		notificationMessage("Watching for file changes on host", "Watch", "")
+
 		stdOut.Printf(
 			"[%s] Watching for file changes on host %s ",
 			green(client.Config.Environment),
@@ -102,6 +109,16 @@ func handleWatchEvent(client kit.ThemeClient, asset kit.Asset, event kit.EventTy
 		blue(asset.Key),
 	)
 	if err := perform(client, asset, event, nil); err != nil {
+		notificationMessage("There is a error in your templates check your logs", "Watch", "")
 		stdErr.Printf("[%s] %s", green(client.Config.Environment), err)
 	}
+}
+
+func notificationMessage(message string, title string, iconPath string) {
+	notify = notificator.New(notificator.Options{
+		DefaultIcon: basepath + "/docs/assets/images/shopify-logo.png",
+		AppName:     "ThemeKit",
+	})
+
+	notify.Push(title, message, basepath + "/docs/assets/images/shopify-logo.png", notificator.UR_NORMAL)
 }
--- a/glide.yaml
+++ b/glide.yaml
@@ -3,6 +3,10 @@ import:
 - package: github.com/fatih/color
   version: ^1.0.0
 - package: github.com/inconshreveable/go-update
+- package: github.com/0xAX/notificator
+  repo: github.com/0xax/notificator
+  vcs: git
+  version: master
 - package: github.com/ryanuber/go-glob
   version: ^0.1.0
 - package: github.com/spf13/cobra


cmd/watch.go
```diff
@@ -102,6 +109,16 @@ func handleWatchEvent(client kit.ThemeClient, asset kit.Asset, event kit.EventTy
 		blue(asset.Key),
 	)
 	if err := perform(client, asset, event, nil); err != nil {
+		notificationMessage("There is a error in your templates check your logs", "Watch", "")
 		stdErr.Printf("[%s] %s", green(client.Config.Environment), err)
 	}
 }
+
+func notificationMessage(message string, title string, iconPath string) {
+	notify = notificator.New(notificator.Options{
```
tanema: I would prefer to have this disabled by default and enabled using a command line flag. I don't think the default behavior of any command line tool should leave that domain unless given permission explicitly. In this case by a command line flag.

cmd/watch.go
```diff
@@ -102,6 +109,16 @@ func handleWatchEvent(client kit.ThemeClient, asset kit.Asset, event kit.EventTy
...
+		DefaultIcon: basepath + "/docs/assets/images/shopify-logo.png",
```
tanema: How would this work with a single binary distribution. Part of the design of themekit is to be a single binary distro to make it easily portable. If notificator could take a file instance we could create a go file with zip data of an icon.

cmd/watch.go
```diff
@@ -102,6 +109,16 @@ func handleWatchEvent(client kit.ThemeClient, asset kit.Asset, event kit.EventTy
...
+		AppName:     "ThemeKit",
```
tanema: I would say this should be "Shopify Themekit" Also there should be a flag to alter this so that tools that want to leverage this like slate can rename it.

cmd/watch.go
```diff
@@ -102,6 +109,16 @@ func handleWatchEvent(client kit.ThemeClient, asset kit.Asset, event kit.EventTy
 		blue(asset.Key),
 	)
 	if err := perform(client, asset, event, nil); err != nil {
+		notificationMessage("There is a error in your templates check your logs", "Watch", "")
 		stdErr.Printf("[%s] %s", green(client.Config.Environment), err)
 	}
 }
+
+func notificationMessage(message string, title string, iconPath string) {
```
tanema: Does this method need an iconPath if you never use it or pass it? Also if this is only used for watch then does it need the title param?

tanema:
Also to run `make check` you will need to install golint by running `go get -u github.com/golang/lint/golint`

darol100:
@tanema, Thanks for the feedback. 

I will work on all those suggestion sometime this week. I knew this PR was not ready, but I just wanted to get some feedback on this behavior.

cmd/watch.go
darol100: In [Laravel Mix](https://github.com/JeffreyWay/laravel-mix), which is where I got the idea they have it enabled by default but we can make into flag that is not an issue.

cmd/watch.go
darol100: The `DefaultIcon` was not even working for me. It was showing the default notification icon.

> How would this work with a single binary distribution

I do not know yet. I will have to investigate your suggestion about zipping the data of the icon.

darol100:
> what platform are you on and what platforms did you test this on?

I test it out on `macOS Sierra Version 10.12.6`

cmd/watch.go
darol100: There is no need to pass the param. I will remove that.

tanema:
Also you should switch to using github.com/martinlindhe/notify for notifications. It uses separate libraries for Windows and OSX to make sure that it is fully supported.

darol100:
@tanema ,

Sorry for the late response, but I am low on time. Feel free to take over the PR.

Thanks.

louiswalch:
I know this is an old post, but has anyone else had success implementing desktop notifications to Shopify Theme Kit's `watch`? I don't use Laravel, and the other repo above seems to be depreciated. It's looks like osX has a native command `osascript` for triggering notifications, but haven't yet had success hooking it into Theme Kit.

PR #499, opened 2017-12-10
Use try-catch to exit PowerShell script on error

ADTC:
### WHAT are you trying to accomplish with this PR?

Use try-catch to exit PowerShell script on error condition (abnormal circumstances).

Script should not continue and report "Install Complete" when there is an error. Instead, it should print out an error message and exit with an exit code indicating failure.

Note that `Write-Output` writes to the pipeline while `Write-Host` writes to the host terminal (usually the user's screen). Only `Write-Host` provides coloring options. _(We should be using `Write-Host` for all messages, but I didn't change that here. I can open another PR for that.)_

### Checklist

- [x] It is safe to simply rollback this change (e.g. changes to `theme.lock` that might corrupt a users workflow).
- [ ] I have :tophat:'d these changes by using the commands I changed by hand. _(what does tophatting mean?)_
- [x] I have run successfully run linting and tests on the codebase. _(script execution was tested)_
- [ ] I have added tests for the changes I have made.
- [ ] I have added a dependency to the project.

_(I'm not sure of the relevancy of some of the checklist items, as I'm changing the PowerShell script only. Please let me know if you need more info.)_

Mentions: @tanema, @chrisbutcher

### DIFF
--- a/scripts/install.ps1
+++ b/scripts/install.ps1
@@ -2,7 +2,14 @@
 $latest_release_url = "https://shopify-themekit.s3.amazonaws.com/releases/latest.json"
 
 Write-Output "Fetching release data";
-$release = $web_client.DownloadString($latest_release_url) | ConvertFrom-Json
+Try {
+  $release = $web_client.DownloadString($latest_release_url) | ConvertFrom-Json
+} Catch {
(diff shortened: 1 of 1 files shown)


tanema:
This a great! Thanks for your help!

PR #589, opened 2018-10-15
Add automatic environment variable loading and Variables in yml config.

tanema:
### Features
This PR encapsulates 2 main features.
- Automatically load environment variables from a file during runtime
- Allow environment variable interpolation in the config 

This will allow developers to much more easily commit their code to public repos without revealing their passwords or their clients.

### Environment variable loading 
The normal user configuration paths are used so depending on OS the path used would be:
Windows:  `%APPDATA%\Shopify\Themekit\variables` 
Linux/BSDs: `${XDG_CONFIG_HOME}/Shopify/Themekit/variables` 
MacOSX: `${HOME}/Library/Application Support/Shopify/Themekit/variables`
    
Themekit will also look for a variables file within the project directory. There has also been a flag added `--vars` will provide a path to a variables file.

The format of these files would look like a normal .env file:
```
DEV_PASSWD=0bwef09hn23048sdkl2345n2k3
```

### Config file interpolation.
The config looks would like this:
```
development:
  password: ${DEV_PASSWD}
  theme_id: ${DEV_THEMEID}
  store: ${DEV_SHOP}
```

and the commands can be run as normal. This will not add any breaking functionality and should not break any existing workflows

### Warn Checklist
- [x] I have added a dependancy to the project.

### DIFF
--- a/changelog.txt
+++ b/changelog.txt
@@ -4,6 +4,9 @@ v1.0.0
 - Fixed a problem with uploading compiled assets
 - Decoupled config file from command execution so any command can run without
 	a config file.
+- Added environment variable interpolation into config files.
+- Added a global environment variable file so that variables can be easily
+managed
 
 v0.8.1 (Sept 18, 2018)
 ======================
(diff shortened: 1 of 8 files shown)


tanema:
use https://github.com/shibukawa/configdir for loading a environment file

PR #629, opened 2019-04-05
[DNM] Experiment for Hackdays to use checksums on our asset APIs

tanema:
This will not work on production shopify as this was done with a local running shopify server and an experimental branch.

### What this does

During deploy, watch and download, it will compare the checksums of the files on
shopify compared with local checksums and will skip files that are the same. It will
also pass the previous checksum in the requests to shopify in the
`checksum_lock` param to ensure the version being updated has not
changed since last observed. Lastly for json files, we compact them
before checksuming since that is how checksums are calculated on
shopify.

### DIFF
--- a/src/shopify/theme_client.go
+++ b/src/shopify/theme_client.go
@@ -38,6 +38,9 @@ var (
 	ErrMissingAssetName = errors.New("asset has no name so could not be processes")
 	// ErrThemeNameRequired is returned when trying to create a theme with a blank name
 	ErrThemeNameRequired = errors.New("theme name is required to create a theme")
+	// ErrContentLock is returned when a modify call is made on an asset with a checksum passed.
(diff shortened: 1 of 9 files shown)


src/shopify/theme_client.go
```diff
@@ -326,6 +341,8 @@ func (c Client) DeleteAsset(asset Asset) error {
 		return ErrNotPartOfTheme
 	} else if resp.StatusCode == 406 {
 		return ErrMissingAssetName
+	} else if resp.StatusCode == 409 {
```
andyw8, line 344: > **409 Conflict**
> Indicates that the request could not be processed because of conflict in the current state of the resource, such as an edit conflict between multiple simultaneous updates.

https://en.wikipedia.org/wiki/List_of_HTTP_status_codes

Precha24 (approved):
Good, thanks

PR #687, opened 2020-02-12
Py3+ compatibility for install script

jnhmcknight:
This allows the installer script to run under either Python 2 or 3, at the user's discretion, or with whatever happens to be installed on the system as the default version. As time goes by, Python 3 will be more commonly the default, and even the only, Python version pre-installed on most systems.

Additionally, it is now also possible to pass the desired install path as an argument when running the script, as the path was already accepted in the installer class, but not able to be passed to it without modifying the installer. Example: `./install.py <path/to/where/i/want/themekit/installed>` Of note, the default install path has not changed and will apply if the user doesn't specify one.

/cc @tanema

### DIFF
--- a/docs/scripts/install.py
+++ b/docs/scripts/install.py
@@ -1,9 +1,9 @@
 #!/usr/bin/env python
 '''
     File name: install
-    Author: Tim Anema
+    Author: Tim Anema, Jared McKnight
     Date created: Sep 29, 2016
-    Date last modified: Sep 14 2018
+    Date last modified: Feb 12 2020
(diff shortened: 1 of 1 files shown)


jnhmcknight:
Guess Travis doesn't like having my name as part of the author comment...

Not really sure what happened with the result at `watcher_test.go:120`. Your Python may be rusty, @tanema, but my Go is non-existent.

Can either drop that non-essential commit or perhaps it was a CI glitch and rerunning Travis will succeed this time. Let me know which you'd prefer.

PR #723, opened 2020-06-24
Skip download of files if checksums match

andyw8:
* Refactored to pass around Assets rather than files paths. This is to make it easier to compare checksums.
* Implemented behaviour to skip downloads where the local checksum matches the server checksum.
* Added some verbose output for easier troubleshooting to indicate skipped files.
* Updated the tests as necessary.
* Fixed an existing bug in the `compileAssetFilenames` method.
* Altered the API call to fetch checksum from server.

Pairing with @tanema, @ayronshopify and @carmelal 
### Warn Checklist

### DIFF
--- a/cmd/download.go
+++ b/cmd/download.go
@@ -46,7 +46,7 @@ func download(ctx *cmdutil.Ctx) error {
 			defer downloadGroup.Done()
 
 			localAsset, _ := shopify.ReadAsset(ctx.Env, requestAsset.Key)
-			if localAsset.Checksum == requestAsset.Checksum {
+			if localAsset.Checksum == requestAsset.Checksum && requestAsset.Checksum != "" {
 				ctx.Log.Printf("[%s] Skipped %s (%s)", colors.Green(ctx.Env.Name), colors.Blue(requestAsset.Key), localAsset.Checksum)
 			} else if asset, err := ctx.Client.GetAsset(requestAsset.Key); err != nil {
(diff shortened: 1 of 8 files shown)


cmd/download.go
```diff
@@ -28,76 +29,75 @@ var downloadCmd = &cobra.Command{
...
 		downloadGroup.Add(1)
-		go func(filename string) {
+		go func(requestAsset shopify.Asset) {
 			defer ctx.DoneTask()
 			defer downloadGroup.Done()
-			if asset, err := ctx.Client.GetAsset(filename); err != nil {
-				ctx.Err("[%s] error downloading %s: %s", colors.Green(ctx.Env.Name), colors.Blue(filename), err)
+
+			localAsset, _ := shopify.ReadAsset(ctx.Env, requestAsset.Key)
+			if localAsset.Checksum == requestAsset.Checksum {
```
andyw8, line 49: @ayronshopify we should probably add a check in case both are nil or blank.

cmd/download_test.go
```diff
@@ -11,7 +11,16 @@ import (
 )
 
 func TestDownload(t *testing.T) {
-	allAssets := []shopify.Asset{{Key: "assets/logo.png"}, {Key: "templates/customers/test.liquid"}, {Key: "config/test.liquid"}, {Key: "layout/test.liquid"}, {Key: "snippets/test.liquid"}, {Key: "templates/test.liquid"}, {Key: "locales/test.liquid"}, {Key: "sections/test.liquid"}}
+	allAssets := []shopify.Asset{
```
andyw8, line 14: Just reformatting.

PR #724, opened 2020-06-29
Skip uploads if checksums match

andyw8:
Co-authored with @ayronshopify 

Depends on this Core change: https://github.com/Shopify/shopify/pull/251785

TODO:

- [ ] squash commits after final approval

This PR adds behaviour so that if a files local checksum matches the remote checksum, then we skip it.

The existing upload process had two variants:
- If no list of files is provided to `deploy`, it will first make a GET request to discover what files are on the server. From this it can determine if any local files should be deleted.
- If a list of files it provided to `deploy`, then it does not make the GET request. 

To compare local checksums against remote checksums, we now need to make the GET request for both these situations.

Note: There are also fixes to no longer show skipped files when in non-verbose mode.

- [x] Verify how `watch` behaves now.

### Warn Checklist
- [ ] ~I have added a dependancy to the project.~ No new dependencies

### DIFF
--- a/cmd/deploy.go
+++ b/cmd/deploy.go
@@ -80,15 +80,17 @@ func deploy(ctx *cmdutil.Ctx) error {
 
 func generateActions(ctx *cmdutil.Ctx) (map[string]file.Op, error) {
 	assetsActions := map[string]file.Op{}
+	pathsToChecksums := map[string]string{}
 
-	if len(ctx.Args) == 0 && !ctx.Flags.NoDelete {
-		remoteFiles, err := ctx.Client.GetAllAssets()
-		if err != nil {
-			return assetsActions, err
+	remoteFiles, err := ctx.Client.GetAllAssets()
+	if err != nil {
+		return assetsActions, err
+	}
+	for _, remoteAsset := range remoteFiles {
+		if len(ctx.Args) == 0 && !ctx.Flags.NoDelete {
+			assetsActions[remoteAsset.Key] = file.Remove
 		}
-		for _, asset := range remoteFiles {
-			assetsActions[asset.Key] = file.Remove
-		}
+		pathsToChecksums[remoteAsset.Key] = remoteAsset.Checksum
 	}
 
 	localAssets, err := shopify.FindAssets(ctx.Env, ctx.Args...)
@@ -103,7 +105,11 @@ func generateActions(ctx *cmdutil.Ctx) (map[string]file.Op, error) {
 
 	for _, asset := range localAssets {
 		var path = asset.Key
-		assetsActions[path] = file.Update
+		if asset.Checksum != "" && (asset.Checksum == pathsToChecksums[asset.Key]) {
+			assetsActions[path] = file.Skip
+		} else {
+			assetsActions[path] = file.Update
+		}
 	}
 	return assetsActions, nil
 }
--- a/cmd/deploy_test.go
+++ b/cmd/deploy_test.go
@@ -14,41 +14,76 @@ import (
 	"github.com/Shopify/themekit/src/shopify"
 )
 
-func TestUpload(t *testing.T) {
+func TestUploadSingleFile(t *testing.T) {
 	ctx, client, _, _, _ := createTestCtx()
 	ctx.Args = []string{"templates/layout.liquid"}
 	ctx.Flags.NoDelete = true
+	client.On("GetAllAssets").Return([]shopify.Asset{{Key: "templates/layout.liquid"}}, nil)
 	client.On("UpdateAsset", shopify.Asset{Key: "templates/layout.liquid"}).Return(nil)
 	err := deploy(ctx)
 	assert.NotNil(t, err)
+}
 
-	ctx, client, _, _, _ = createTestCtx()
+func TestUploadWithReadOnlyOption(t *testing.T) {
+	ctx, client, _, _, _ := createTestCtx()
 	ctx.Args = []string{"templates/layout.liquid"}
 	ctx.Flags.NoDelete = true
 	ctx.Env.ReadOnly = true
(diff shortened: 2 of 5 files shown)


cmd/watch.go
```diff
@@ -79,6 +79,14 @@ func watch(ctx *cmdutil.Ctx, events chan file.Event, sig chan os.Signal) error {
 func perform(ctx *cmdutil.Ctx, path string, op file.Op) {
 	defer ctx.DoneTask()
 
+	if op == file.Skip {
+		// TODO: which colour for Skipped files?
```
andyw8, line 83: @attaboy Any thoughts here? Do we need to distinguish skipped files from non-skipped?

cmd/deploy_test.go
```diff
@@ -14,45 +14,66 @@ import (
 	"github.com/Shopify/themekit/src/shopify"
 )
 
-func TestUpload(t *testing.T) {
+func TestUploadSingleFile(t *testing.T) {
```
andyw8, line 17: @tanema The existing test was getting pretty long, is it ok to split it up like this?

src/file/watcher.go
```diff
@@ -18,6 +18,8 @@ const (
 	Update Op = iota
 	// Remove is a file op where the file is removed
 	Remove
+	// Skip is file op where the remote file matches the local file so is not transferred
+	Skip
```
andyw8, line 22: @tanema Does this seem a reasonable approach?

cmd/deploy_test.go
```diff
@@ -14,45 +14,66 @@ import (
...
+	client.On("GetAllAssets").Return([]shopify.Asset{{Key: "assets/app.js"}}, nil)
 	client.On("UpdateAsset", shopify.Asset{Key: "assets/app.js", Checksum: "d41d8cd98f00b204e9800998ecf8427e"}).Return(nil)
-	err = deploy(ctx)
+	err := deploy(ctx)
 	assert.Nil(t, err)
 	assert.Contains(t, stdOut.String(), "Updated assets/app.js")
+}
 
-	ctx, client, _, stdOut, _ = createTestCtx()
+func TestUploadEverything(t *testing.T) {
```
ayronshopify, line 53: I would maybe call this `TestUploadAllFiles` or `TestUploadWithoutArgs`

src/file/watcher.go
tanema: Yeah I think it is great and could be used in the watcher as well.

cmd/deploy_test.go
andyw8: Fixed in 7e0195a

cmd/deploy_test.go
andyw8: Fixed in 7e0195a

cmd/watch.go
andyw8: This will be addressed in https://github.com/Shopify/themekit/issues/727

cmd/deploy.go
```diff
@@ -80,15 +80,17 @@ func deploy(ctx *cmdutil.Ctx) error {
 
 func generateActions(ctx *cmdutil.Ctx) (map[string]file.Op, error) {
 	assetsActions := map[string]file.Op{}
+	pathToChecksums := map[string]string{}
 
-	if len(ctx.Args) == 0 && !ctx.Flags.NoDelete {
```
tanema, line 84: It looks like you have taken out this default case? This is the case where no filenames are provided, does theme deploy work if you just pass a single filename? or multiple?

cmd/download.go
```diff
@@ -47,7 +47,9 @@ func download(ctx *cmdutil.Ctx) error {
 
 			localAsset, _ := shopify.ReadAsset(ctx.Env, requestAsset.Key)
 			if localAsset.Checksum == requestAsset.Checksum && requestAsset.Checksum != "" {
-				ctx.Log.Printf("[%s] Skipped %s (%s)", colors.Green(ctx.Env.Name), colors.Blue(requestAsset.Key), localAsset.Checksum)
+				if ctx.Flags.Verbose {
```
tanema, line 50: So it looks like this PR includes changes to download as well, you should include that in the PR description.

cmd/watch.go
```diff
@@ -79,6 +79,13 @@ func watch(ctx *cmdutil.Ctx, events chan file.Event, sig chan os.Signal) error {
 func perform(ctx *cmdutil.Ctx, path string, op file.Op) {
 	defer ctx.DoneTask()
 
+	if op == file.Skip {
+		localAsset, _ := shopify.ReadAsset(ctx.Env, path)
+		checksumOutput := "Checksum: " + localAsset.Checksum
+		ctx.Log.Printf("[%s] Skipping %s (%s)", colors.Green(ctx.Env.Name), colors.Blue(path), checksumOutput)
+		return
+	}
+
```
tanema, line 88: The remove if statement could just be an `else if` now and then you don't need that return.

andyw8: How about using `case`? 9c586d2

cmd/deploy.go
andyw8: The behaviour is still the same. We had to change it since the `GetAllAssets` call to the server now needs to happen on the _no-args_ and _with-args_ variants of `deploy`, whereas before it only happened on the _no-args_ variant.

cmd/download.go
andyw8: I also committed 96344d7 for a similar purpose of hiding the skips when not in verbose mode.

cmd/watch.go
tanema: Sure, looks good!

cmd/deploy.go
tanema: haha, I was there for that change 🤦

PR #730, opened 2020-07-14
feat: publish a docker image each time a GitHub release is generated 

thisislawatts:
### Warn Checklist
- [X] I have added a dependancy to the project.

I am looking to use themekit as part of our CI/CD pipeline. Rather than introducing Python as a requirement for the build process it would be great if there was a Docker image I could pull down which already has `themekit` available.

Adding more manual processes to a Release process if no way to make friends, so this PR introduces a GitHub workflow to automatically publish to DockerHub each time a Release is created on GitHub.

All that you need to do is add DockerHub username and password to the GitHub repository.

```
docker run --rm thisislawatts/themekit version
```
Note: The above command will output 1.0.6 as [I incremented the version in my forked repo during testing](https://github.com/thisislawatts/themekit/commit/72cf456305d2c4cf92f46b6c5f2b7ca8aeaba5af)

Look forward to hearing your thoughts.

### DIFF
--- a/.github/workflows/dockerfile-release.yml
+++ b/.github/workflows/dockerfile-release.yml
@@ -15,4 +15,5 @@ jobs:
           username: ${{ secrets.DOCKER_USERNAME }}
           password: ${{ secrets.DOCKER_PASSWORD }}
           repository: thisislawatts/themekit
+          tags: latest
           tag_with_ref: true
--- a/.github/workflows/dockerfile-release.yml
+++ b/.github/workflows/dockerfile-release.yml
@@ -1,0 +1,18 @@
+name: Publish Docker image
+on:
+  release:
+    types: [published]
+jobs:
+  push_to_registry:
+    name: Push Docker image to Docker Hub
+    runs-on: ubuntu-latest
+    steps:
+      - name: Check out the repo
+        uses: actions/checkout@v2
+      - name: Push to Docker Hub
+        uses: docker/build-push-action@v1
+        with:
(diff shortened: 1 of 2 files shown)


.github/workflows/dockerfile-release.yml
```diff
@@ -0,0 +1,19 @@
...
+    runs-on: ubuntu-latest
+    steps:
+      - name: Check out the repo
+        uses: actions/checkout@v2
+      - name: Push to Docker Hub
+        uses: docker/build-push-action@v1
+        with:
+          username: ${{ secrets.DOCKER_USERNAME }}
+          password: ${{ secrets.DOCKER_PASSWORD }}
+          repository: thisislawatts/themekit
```
thisislawatts, line 17: This must be changed to match the DockerHub repository you setup.

thisislawatts:
I have signed the CLA but the status check seems to disagree. Here's evidence I when trying to resign the document:
screenshot 2020-07-14 at 21 23 01

tanema:
Thanks @thisislawatts I will have to come back to this soon because I need to figure out shopifys docker setup and if we can do this.

thisislawatts:
Thanks for getting back to me @tanema, much appreciated. I wasn't sure of the current state of [Shopify on DockerHub](https://hub.docker.com/u/shopify) as the public profile hasn't had any new images added in 10 months.

thisislawatts:
Hey @tanema, wondering if there is anything I can help with here?

thisislawatts:
Closing due to lack of engagement from maintainers

PR #733, opened 2020-07-20
Minor UX tweaks

### LINKED ISSUE #728: [UX] Displaying diagnostics
As part of the checksums work, we are proposing adding some diagnostics to display checksums when using themekit in verbose mode.

Displaying the checksums is not strictly needed, but I believe it will be a useful tool for troubleshooting any issues that might come up, since we can ask users to share the output with us.

We display the remote checksum, followed by the local checksum, e.g.:

`[development] Successfully wrote locales/en.default.json to disk (Remote Checksum faf7f6489cc076a27b9a606e7d37308f, Local Checksum cca038c70f1ee2cd5ececed6c8216901)`

In some situations the remote checksum may not exist yet (but the local checksum would always be available).

Some things to consider:

- Could the checksum 'noise' be annoying for users?
- Should the checksums only be shown if we supply a particular flag?
- Should it displayed in a different colour?

@Nupur-Mukherjee @attaboy @tanema

andyw8:
As agreed with @Nupur-Mukherjee .

### Warn Checklist

Fixes #728

### DIFF
--- a/cmd/deploy_test.go
+++ b/cmd/deploy_test.go
@@ -72,7 +72,7 @@ func TestUploadForSkipFileWhenChecksumsMatch(t *testing.T) {
 	client.On("UpdateAsset", shopify.Asset{Key: "assets/app.js", Checksum: "d41d8cd98f00b204e9800998ecf8427e"}).Return(nil)
 	err := deploy(ctx)
 	assert.Nil(t, err)
-	assert.Contains(t, stdOut.String(), "Skipping config/settings_data.json")
+	assert.Contains(t, stdOut.String(), "Skipped config/settings_data.json")
 	assert.Contains(t, stdOut.String(), "Updated assets/app.js")
 }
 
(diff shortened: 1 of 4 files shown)


cmd/deploy_test.go
```diff
@@ -72,7 +72,7 @@ func TestUploadForSkipFileWhenChecksumsMatch(t *testing.T) {
 	client.On("UpdateAsset", shopify.Asset{Key: "assets/app.js", Checksum: "d41d8cd98f00b204e9800998ecf8427e"}).Return(nil)
 	err := deploy(ctx)
 	assert.Nil(t, err)
-	assert.Contains(t, stdOut.String(), "Skipping config/settings_data.json")
+	assert.Contains(t, stdOut.String(), "Skipped config/settings_data.json")
```
andyw8, line 75: For consistency.

src/colors/colors.go
```diff
@@ -20,4 +20,6 @@ var (
 	ColorStdOut = log.New(colorable.NewColorableStdout(), "", 0)
 	// ColorStdErr is a wrapped std err that allows colors
 	ColorStdErr = log.New(colorable.NewColorableStderr(), "", 0)
+	// BrightBlack is the color Black
```
andyw8, line 23: (Not really a helpful comment, but it won't compile without something in here).

PR #745, opened 2020-07-30
Send previous checksum with updates

andyw8:
This adds new behaviour for the `watch` command. We send a new header `X-Shopify-Replace-If-Checksum-Match` when updating an asset. This allows us to determine if an asset has been changed by someone else working on the site at the same time, to avoid accidental overwrites.

The server-side aspect of this has not yet been implemented, but this is a step towards that.

### Warn Checklist


src/shopify/theme_client.go
```diff
@@ -314,6 +314,41 @@ func (c Client) UpdateAsset(asset Asset) error {
 	return nil
 }
 
+// UpdateWatchedAsset will take an asset and will return when the asset has been updated.
+// If there was an error, in the request then error will be defined otherwise the
+//response will have the appropropriate data for usage.
+func (c Client) UpdateWatchedAsset(asset Asset, oldChecksum string) error {
```
andyw8, line 320: Currently this heavily duplicates the existing `UpdateAsset` but I should be able to extract out the commonality.

cmd/watch.go
```diff
@@ -52,12 +52,12 @@ var watchCmd = &cobra.Command{
 			defer watcher.Stop()
 			signalChan := make(chan os.Signal)
 			signal.Notify(signalChan, os.Interrupt)
-			return watch(ctx, watcher.Events, signalChan)
+			return watch(ctx, watcher.Events, signalChan, checksums)
```
andyw8, line 55: This is the map of paths to checksums as fetched from the server upon running `watch`.

cmd/watch.go
```diff
@@ -79,14 +79,14 @@ func watch(ctx *cmdutil.Ctx, events chan file.Event, sig chan os.Signal) error {
 				return cmdutil.ErrReload
 			}
 			ctx.Log.Printf("[%s] processing %s", colors.Green(ctx.Env.Name), colors.Blue(event.Path))
-			perform(ctx, event.Path, event.Op)
+			perform(ctx, event.Path, event.Op, checksums)
```
andyw8, line 82: Maybe this can just be the individual checksum rather than the whole map?

src/file/watcher.go
```diff
@@ -40,6 +40,7 @@ var (
 type Event struct {
 	Op       Op
 	Path     string
+	OldChecksum string
```
andyw8, line 43: Maybe `ServerChecksum` or `RemoteChecksum`?

andyw8:
@tanema I'm still doing some testing and cleanup, but let me know if you have any initial thoughts here.

src/shopify/theme_client.go
tanema: Why not just update the `UpdateAsset` method call and only add the header if it is non-empty.

src/file/watcher.go
tanema: I think `PreviousChecksum` might be better because the previous checksum might come from a lockfile in the future.

cmd/watch.go
tanema: Why pass in a map? We could pass in the previous checksum here from the event right?

src/shopify/theme_client.go
```diff
@@ -280,14 +280,18 @@ func (c Client) GetAsset(filename string) (Asset, error) {
...
-// UpdateAsset will take an asset and will return  when the asset has been updated.
+// UpdateAsset will take an asset and will return when the asset has been updated.
 // If there was an error, in the request then error will be defined otherwise the
 // response will have the appropriate data for usage.
-func (c Client) UpdateAsset(asset Asset) error {
-	resp, err := c.http.Put(c.assetPath(map[string]string{}), map[string]Asset{"asset": asset}, nil)
+func (c Client) UpdateAsset(asset Asset, prevChecksum string) error {
+	var header = make(map[string]string)
+	if prevChecksum != "" {
+		header["X-Shopify-Previous-Checksum"] = prevChecksum
```
andyw8, line 292: @attaboy @gfscott 

@tanema suggested getting your thoughts on what to call this header (`X-Shopify-Something`).

It's for detecting concurrency conflicts, and it somewhat similar to an [Etag](https://en.wikipedia.org/wiki/HTTP_ETag).

When`theme watch` is started, we grab the checksums of the assets on the server. When an asset is updated, we send the old checksum as part of the `PUT`. We can then determine if the asset was changed another user, and avoid accidental overwrites.

'previous' seems a little vague... any thoughts on something clearer?

gfscott: Interesting! Agree that `previous` isn't quite specific enough.

What about `X-Shopify-Latest-Checksum` — seems like it more accurately identifies that this is "the single most recent checksum" (as opposed to "any previous checksum"). BUT does it risk being confused with the _new_ checksum of the updated asset?

andyw8: `X-Shopify-Last-Known-Checksum` maybe?

andyw8: or `X-Shopify-Checksum-On-Download`

gfscott: I like `Last-Known-Checksum`, I think it's properly specific and unambiguous.

andyw8: 👍 Updated in b5e12a5

andyw8: @gfscott, @tanema suggested using something closer to the `If-Match` convention used be ETags:

https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/ETag

We're now thinking `Replace-If-Checksum-Match`, which is like saying to the server, "I'm giving this updated asset to you, if the checksum it's signed with matches your curret checksum, then replace it, otherwise reject it".

gfscott: Cool! if there's a recognizable pattern we can lean on, that's preferable for sure.

Does the development handbook have general guidelines on naming for things like http headers? I did [a quick search](https://development.shopify.io/search?q=naming) initially and didn't find much.

PR #747, opened 2020-07-30
Adding Retries and improved throttling to HTTP requests

### LINKED ISSUE #665: Timeout Issue
**Describe the bug**
Hi, when I type `theme watch`, I ran into some issues. I can't track the files. This is what I get: 
`request timed out. if you are receive this error consistently, try increasing the timeout in your config`

Hope you can help me. 
Thanks!

**Environment :**
 - OS [macOS Mojave]:
- Themekit version [1.0.1]:

### LINKED ISSUE #715: Make Deploys Faster and Smarter
Every time `theme deploy` is ran, all files of the the theme are uploaded. However, often only a few files actually changed. This is not only slow and inefficient but also error prone, as Shopifys API request limits are hit often, when a theme is deployed to multiple origins concurrently. 

It would by great if `theme deploy` only deploys and updates those files, that are created, updated or deleted. Also the command should be more resilient to API request limits. Instead of exiting the command when the request limit is hit, the command could pause an retry to upload a file.

**Describe alternatives you've considered**
For now we we are restarting `theme deploy` in our CICD pipeline every time the deploy failed due to hitting the API request limit.

tanema:
fixes #715
fixes #665

### What is this PR solving
This will solve a few errors with asset requests, more specifically timeouts but also hopefully speed up requests during actions like `deploy` especially for plus merchants. 

#### Before
Themekit was throttled directly to 2 requests per second to just try to ensure that a 429 was never received. 

### Now
I have added strategy to throtting.
- Changes to the rate limiter
  - made it nicer to specify requests per second (for myself)
  - using a (x)-std lib library for throttling instead to get the bonus of bursts
  - added `ResetAfter` to halt everything until a future time.
- Changes to the http client
  - rate limiter is not limiting to 4 requests per second (the plus limit)
  - if we get a 429 we record the time that we should resume and the request is set to retry
  - all requests are held until it is that time.
  - once that time has passed, we resume our normal throttling
  - if a timeout is received, it will be retried
  - all timeout requests are retried at a maximum of 5 times
  - if a timeout is received after 5 retries, the user is notified that there may be an issue with their connection.

#### Performance notes
I have noticed that this works noticably faster now but also has the side effect of being really fast at the begining and slower at the end, which might make it feel slower. 

This has hopefully eliminated timeouts,

### Warn Checklist
- [x] I have added a dependancy to the project.

### DIFF
--- a/src/httpify/client.go
+++ b/src/httpify/client.go
@@ -6,11 +6,12 @@ import (
 	"encoding/json"
 	"errors"
 	"fmt"
-	"io"
+	"io/ioutil"
 	"net"
 	"net/http"
 	"net/url"
 	"runtime"
+	"strconv"
 	"strings"
 	"time"
 
@@ -19,7 +20,6 @@ import (
 )
 
 var (
-	errClientTimeout   = errors.New(`request timed out. if you are receive this error consistently, try increasing the timeout in your config`)
 	errConnectionIssue = errors.New("DNS problem while connecting to Shopify, this indicates a problem with your internet connection")
 )
 
@@ -29,7 +29,6 @@ type Params struct {
 	Password string
 	Proxy    string
 	Timeout  time.Duration
-	APILimit time.Duration
 }
 
 // HTTPClient encapsulates an authenticate http client to issue theme requests
@@ -40,6 +39,7 @@ type HTTPClient struct {
 	baseURL  *url.URL
 	client   *http.Client
 	limit    *ratelimiter.Limiter
+	maxRetry int
 }
 
 // NewClient will create a new authenticated http client that will communicate
@@ -60,7 +60,8 @@ func NewClient(params Params) (*HTTPClient, error) {
 		password: params.Password,
 		baseURL:  baseURL,
 		client:   adapter,
-		limit:    ratelimiter.New(params.Domain, params.APILimit),
+		limit:    ratelimiter.New(params.Domain, 4),
+		maxRetry: 5,
 	}, nil
 }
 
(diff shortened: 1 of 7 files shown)


scottdixon:
🎉 This is awesome @tanema. Just tested on a Plus store, downloading the Venture theme (166 files).

Before: 87 seconds
After: 42 seconds

src/httpify/client.go
```diff
@@ -110,14 +102,38 @@ func (client *HTTPClient) do(method, path string, body interface{}, headers map[
...
+			data, err := json.Marshal(body)
+			if err != nil {
+				return nil, err
+			}
+			req.Body = ioutil.NopCloser(bytes.NewBuffer(data))
+		}
+
+		client.limit.Wait()
+		resp, err := client.client.Do(req)
+		if err == nil && resp.StatusCode >= 100 && resp.StatusCode <= 428 {
```
andyw8, line 122: Why do we choose 428 as the cutoff here?

src/httpify/client_test.go
```diff
@@ -99,25 +98,49 @@ func TestClient_do(t *testing.T) {
...
 	client.baseURL.Scheme = "http"
 
 	assert.NotNil(t, client)
 	assert.Nil(t, err)
 
 	_, err = client.do("POST", "/assets.json", body, nil)
-	if assert.NotNil(t, err) {
-		assert.Equal(t, err, errClientTimeout)
-	}
+	assert.EqualError(t, err, "request timed out after 1 retries, there may be an issue with your connection")
```
andyw8, line 142: Will the end user see this message? I can't tell where it gets outputted (and should it have a newline?).

src/httpify/client.go
tanema: Because 429 is too many requests and I need that response for the `retry-after` header.

src/httpify/client_test.go
tanema: Yes the user will see this message if the request times out (5) times, the error is built with fmt.Errorf

tanema: https://github.com/Shopify/themekit/pull/747/files#diff-2d68343eeed320d6dcac7d933fba1f00R127

src/httpify/client.go
andyw8: For clarity, for perhaps we can define a constant such as `HTTPTooManyRequests` and write this as `resp.StatusCode <  HTTPTooManyRequests`

PR #750, opened 2020-08-10
add --notify-upload flag to notify when a file has finished uploading (fixes #749)

### LINKED ISSUE #749: --notify flag does not work as expected and fires before file has finished uploading
**Describe the bug**
I'm using a custom *livereload* implementation to automatically reload the browser once a file has been uploaded to Shopify.

I'm running themekit with `theme watch --notify=/tmp/theme.update` and a livereload script that watches `theme.update`, then refreshes the page once it detects a change in that file. It isn't working as expected and I still need to add a delay for it work. I added some logging to my app and noticed that themekit is touching the theme.update file *before* finishing uploading the file to Shopify servers.

Here's the output of my app (livereload is watching for all changes for logging purposes but only reloading after changes to *theme.update* are detected):
```
[livereload] change detected: assets/bundle.js
[themekit] processing assets/bundle.js
[livereload] change detected: tmp/theme.update
[livereload] Notifying browser to reload [ 'assets/bundle.js' ]
[themekit] Updated assets/bundle.js
```
As you can see, themekit is updating theme.update before the upload is finished.

**To Reproduce**
1. run `theme watch --notify=/tmp/theme.update`
2. Change a file in your theme.
3. *theme.update* will be touched before the file has finished uploading to Shopify servers.

**Expected behavior**
*theme.update*  should be updated once all files have finished uploading to Shopify servers.

felipebrahm:
This fixes #749.

The `--notify` command currently touches a file when all workers are idle. This is not useful for *livereload* purposes.

I added a `--notify-upload` command that will touch a file after a file has finished uploading.

You might want to just completely replace the `--notify` flag with this behavior, but I wasn't sure if you implemented it like that for some particular use case I'm not aware of.

This needs automated tests (I'm already using this in my local project and it works great though) and the new command would need to be added to the docs.

In the future, a really nice addition would be to actually be able to be notified about which files where just uploaded. This could be done via a POST request to a webhook specified by the user.

Disclaimer: this is my first time writing any *go* code. I just quickly added these changes because I started developing a Shopify theme and wanted this change, so I wrote this for myself and then figured I could just submit a PR and let you guys decide if you want to merge it or take it as inspiration to implement my changes in a better way.

### DIFF
--- a/cmd/theme.go
+++ b/cmd/theme.go
@@ -91,6 +91,7 @@ func init() {
 	ThemeCmd.PersistentFlags().BoolVar(&flags.AllowLive, "allow-live", false, "Will allow themekit to make changes to the live theme on the store.")
 
 	watchCmd.Flags().StringVarP(&flags.NotifyFile, "notify", "n", "", "file to touch when workers have gone idle")
+	watchCmd.Flags().StringVarP(&flags.NotifyUploadFile, "notify-upload", "u", "", "file to touch when files have finished uploading")
 	watchCmd.Flags().BoolVarP(&flags.AllEnvs, "allenvs", "a", false, "run command with all environments")
 	removeCmd.Flags().BoolVarP(&flags.AllEnvs, "allenvs", "a", false, "run command with all environments")
 	openCmd.Flags().BoolVarP(&flags.AllEnvs, "allenvs", "a", false, "run command with all environments")
--- a/cmd/watch.go
+++ b/cmd/watch.go
@@ -5,6 +5,7 @@ import (
 	"log"
 	"os"
 	"os/signal"
+	"time"
 
 	"github.com/spf13/cobra"
 
@@ -117,6 +118,10 @@ func perform(ctx *cmdutil.Ctx, path string, op file.Op) {
 			ctx.Err("[%s] (%s) %s", colors.Green(ctx.Env.Name), colors.Blue(asset.Key), err)
 		} else if ctx.Flags.Verbose {
 			ctx.Log.Printf("[%s] Updated %s", colors.Green(ctx.Env.Name), colors.Blue(asset.Key))
+			if ctx.Flags.NotifyUploadFile != "" {
+				os.Create(ctx.Env.NotifyUpload)
+				os.Chtimes(ctx.Env.NotifyUpload, time.Now(), time.Now())
+			}
 		}
 	}
 }
--- a/src/cmdutil/util.go
+++ b/src/cmdutil/util.go
@@ -44,6 +44,7 @@ type Flags struct {
 	Ignores               []string
 	DisableIgnore         bool
 	NotifyFile            string
+	NotifyUploadFile      string
 	AllEnvs               bool
 	Version               string
 	Prefix                string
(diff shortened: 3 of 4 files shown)


tanema:
Hey! Thanks for the change! You are a great opportunity to give some feedback on this change since you are using it specifically for your own needs.

The reason I have never added this is that just because a single file is finished uploading, this does not mean that all changes are finished uploading. So while you are currently using this, how do you find that it performs? Do you find that it touches the notification file at the appropriate times even though you change multiple files?

My other worry is that I am not sure I like that there are 2 different notification flags. If we were to add this one, we would have to decide which one is more important and only keep that one to avoid confusion.

felipebrahm:
In a real world development scenario where you need to be notified of changes in real time you're probably not modifying multiple files at the same time. You're either modifying a CSS bundle or a JS bundle or just one image. If you're modifying multiple files at the same time you might as well just manually reload your browser.

The current state of the `--notify` flag makes it useless, at least for local development live reload purposes. I just added a new flag instead of modifying the existent one because I thought maybe it was implemented that way because of a use case I'm not aware of, but it seem like that's not the case. If you *really* wanted to notify only after all files have finished uploading then instead of checking for a worker idle state, you should check that all workers are done uploading. I don't know enough golang to make this change myself though, and I think notifying after each file has finished uploading is good enough for live reload purposes (and you might argue you actually want to be notified after each file upload and not wait until the largest file is done).

I tried looking online for solutions to this and see how people are implementing live reload for Shopify theme development but all blog posts I could find just said to watch for file changes and then add a delay of X milliseconds before you trigger the live reload, but that means I'm wasting development time. It also means I have to set the delay high enough to handle the largest of my files (worst case scenario).

**If you agree with the changes I made then I could update this PR and just remove my new flag and modify the existent `--notify` flag to have this behaviour.**

Here's a GIF of my current set up, the CSS is being refreshed live without a page reload as soon as the new file is available on Shopify servers (this works for JS too, but it just does a whole page reload then):

themekit-livereload

Here's the live reload script I wrote (the one that prints the `[livereload]` logs in the GIF above). I then used [this Chrome extension](https://chrome.google.com/webstore/detail/livereload/jnihajbhpnppcggbcgedagnkighmdlei?hl=en) to inject the very common `livereload.js` file into my page's frontend.

```javascript
const chokidar = require('chokidar');
const tinylr = require('tiny-lr');

const LIVE_RELOAD_PORT = 35729;

// Create a live reload server instance
const lrserver = tinylr();

lrserver.listen(LIVE_RELOAD_PORT, (err) => {
  console.log(`Live Reload Server Started on port ${LIVE_RELOAD_PORT}`);
});

// Watch the theme directories (to know which files were changed)
// and the theme.update file (to be notified when themekit has finished uploading changes to Shopify)
const filesToWatch = [
  './assets',
  './config',
  './layout',
  './locales',
  './sections',
  './snippets',
  './templates',
  './tmp/theme.update',
];

const watcher = chokidar.watch(filesToWatch, {
  ignoreInitial: true,
});

let filesChanged = [];

watcher.on('all', (event, path) => {
  if(!/^(add|change)$/.test(event)) {
    // If the event is not "add" or "change" then ignore.
    return;
  }

  // If the file is not theme.update...
  if(!/(^|\/)theme\.update$/.test(path)) {
    console.log(`Change detected: ${path}`);
    // We store which file was changed and we wait to notify the browser once theme.update has changed.
    if(!filesChanged.includes(path)) {
      filesChanged.push(path);
    }
    return;
  }

  console.log(`Notify change detected: ${path}`);

  // If the theme.update file changed, then themekit has finished uploading changes to Shopify, so we tell the browser to reload.
  console.log('Notifying browser to reload', filesChanged);
  notifyBrowser(filesChanged);
  filesChanged = [];
});

const notifyBrowser = (files) => {
  lrserver.changed({
    body: { files },
  });
};
```

felipebrahm:
After writing that last comment I decided to go ahead and try the webhook approach I mentioned and it was actually easier to implement than I thought. The code is this other branch: https://github.com/felipebrahm/themekit/commit/538de9ddfab3f41b1345082effed633e2f11e62d

With the webhook approach then I don't need any custom script (like the one I posted in the previous comment) and instead I can just run a common live reload server and then run themekit with `theme watch --env=development --notify-webhook="http://localhost:35729/changed"` and everything just works :)

Let's finish discussing the changes in this PR and *when* the notify file should be touched. After that I could submit a separate PR to add a `--notify-webhook` flag and discuss that separately.

felipebrahm:
Closing this PR, as it's being implemented with proper tests and all in #757

PR #755, opened 2020-08-19
Use unstable API version for checksums support

andyw8:
We expect checksums to be part of the 2020-10 release. Until then, we'll need to point to the unstable version of the API.

### Warn Checklist

### DIFF
--- a/src/shopify/theme_client.go
+++ b/src/shopify/theme_client.go
@@ -143,7 +143,7 @@ func (c Client) GetShop() (Shop, error) {
 
 // Themes will return all the available themes on a domain.
 func (c Client) Themes() ([]Theme, error) {
-	resp, err := c.http.Get("/admin/themes.json", nil)
+	resp, err := c.http.Get("/admin/api/unstable/themes.json", nil)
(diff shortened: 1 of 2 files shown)


src/shopify/theme_client.go
```diff
@@ -344,7 +344,7 @@ func (c Client) DeleteAsset(asset Asset) error {
 func (c Client) assetPath(query map[string]string) string {
 	formatted := "/admin/assets.json"
```
andyw8, line 345: Missed this and and a couple more, updating...

andyw8:
@tanema 👍 done in c072b67

src/shopify/theme_client_test.go
```diff
@@ -16,6 +16,8 @@ import (
 
 var NoHeaders = map[string]string(nil)
 
+// APIPath is the version of the Admin REST API to use
```
tanema, line 19: Does this comment need to be here?

andyw8: Yeah, the linter complains otherwise 😞

PR #757, opened 2020-08-20
Changing notification system

tanema:
This implements @felipebrahm proposal from last week.

#### What this is doing
- removed notification functionality from the file watcher
- added the notification functionality to come after perform operations in the watch command
- implemented an adapter pattern for 
  - noop: no notify setting
  - url: for url as the notify setting
  - file: for a filepath as the notify setting
- Pass in a notification adapter to the watch command depending on the notification setting.

#### References:
- notify after http call https://github.com/Shopify/themekit/pull/750
- notify webhook https://github.com/felipebrahm/themekit/commit/538de9ddfab3f41b1345082effed633e2f11e62d
- Issue https://github.com/Shopify/themekit/issues/749

### Warn Checklist

### DIFF
--- a/cmd/notify_adapters.go
+++ b/cmd/notify_adapters.go
@@ -1,0 +1,64 @@
+package cmd
+
+import (
+	"bytes"
+	"encoding/json"
+	"net/http"
+	"net/url"
+	"os"
+	"time"
+
+	"github.com/Shopify/themekit/src/cmdutil"
+	"github.com/Shopify/themekit/src/colors"
+)
+
+type notifyAdapter interface {
+	notify(*cmdutil.Ctx, string)
+}
+
+func newNotifyAdapter(notifyPath string) notifyAdapter {
+	if notifyPath == "" {
+		return &noopNotify{}
+	} else if u, err := url.Parse(notifyPath); err == nil && u.Scheme != "" && u.Host != "" {
+		return &urlNotify{
+			url: notifyPath,
+			client: http.Client{
+				Timeout: time.Second,
+			},
+		}
+	}
+	return &fileNotify{path: notifyPath}
+}
+
+type noopNotify struct{}
+
+func (noop *noopNotify) notify(*cmdutil.Ctx, string) {}
+
+type urlNotify struct {
+	url    string
+	client http.Client
+}
+
+func (urlNote *urlNotify) notify(ctx *cmdutil.Ctx, path string) {
+	body, _ := json.Marshal(map[string]interface{}{"files": []string{path}})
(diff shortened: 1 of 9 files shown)


cmd/notify_adapters.go
```diff
@@ -0,0 +1,64 @@
...
+func (urlNote *urlNotify) notify(ctx *cmdutil.Ctx, path string) {
+	body, _ := json.Marshal(map[string]interface{}{"files": []string{path}})
+	resp, err := urlNote.client.Post(urlNote.url, "application/json", bytes.NewBuffer(body))
+	if err != nil {
+		ctx.Log.Printf(
+			`[%s] Error while notifying webhook "%s": %s`,
+			colors.Green(ctx.Env.Name),
+			colors.Blue(ctx.Env.Notify),
+			err,
+		)
```
andyw8, line 51: Does `resp.Body` need to be closed here too to avoid any connecting hanging around?

cmd/theme.go
```diff
@@ -90,7 +90,7 @@ func init() {
 	ThemeCmd.PersistentFlags().BoolVar(&flags.DisableIgnore, "no-ignore", false, "Will disable config ignores so that all files can be changed")
 	ThemeCmd.PersistentFlags().BoolVar(&flags.AllowLive, "allow-live", false, "Will allow themekit to make changes to the live theme on the store.")
 
-	watchCmd.Flags().StringVarP(&flags.NotifyFile, "notify", "n", "", "file to touch when workers have gone idle")
+	watchCmd.Flags().StringVarP(&flags.Notify, "notify", "n", "", "file to touch when workers have gone idle")
```
andyw8, line 93: I guess this description needs updated?

cmd/watch.go
```diff
@@ -80,6 +83,9 @@ func watch(ctx *cmdutil.Ctx, events chan file.Event, sig chan os.Signal) error {
 			}
 			ctx.Log.Printf("[%s] processing %s", colors.Green(ctx.Env.Name), colors.Blue(event.Path))
 			perform(ctx, event.Path, event.Op, event.LastKnownChecksum)
+			if event.Op != file.Skip && event.Op != file.Remove {
```
andyw8, line 86: Might we still want `remove` to trigger a notification? e.g. if you delete an image that seems to be used, you may want to still reload the page to ensure nothing has broken.

cmd/notify_adapters.go
tanema: If there is an error, resp.Body is normally nil and will panic if you try to close it.

PR #760, opened 2020-08-25
Automatically allowing live on open and download

tanema:
We got some feedback from beta testing that the `--allow-live` flag is not really needed on the open and download commands so this PR automatically adds the flag to those commands so that they are able to do these actions.

#### Addition
It also occurred to me that it would make things nicer if we added output about the `live` config from the past, so I added a specific error for a live configuration.

### Warn Checklist


andyw8:
FYI @liamgriffin

tanema:
Just as a note, this will still warn the user that the theme is live during these operations

PR #771, opened 2020-09-01
Update release guidance

andyw8:


andyw8:
(Ignoring intermittent failing tests since this only affect the docs).

Issue #790, opened 2020-09-11
Crash when checking response status code


andyw8:
We received a crash report, seems that we need to check that `resp` is not null before we check _`resp.StatusCode`:

```
panic: runtime error: invalid memory address or nil pointer dereference
[signal SIGSEGV: segmentation violation code=0x1 addr=0x10 pc=0x145c8f2]
goroutine 78 [running]:
github.com/Shopify/themekit/src/httpify.(*HTTPClient).doWithRetry(0xc0002b2800, 0xc00016e000, 0x14ee800, 0xc0000a8c30, 0x1, 0xc00034a190, 0x0)
        /Users/andyw8/src/github.com/Shopify/themekit/src/httpify/client.go:130 +0x102
github.com/Shopify/themekit/src/httpify.(*HTTPClient).do(0xc0002b2800, 0x15775a6, 0x3, 0xc0002cba00, 0x33, 0x14ee800, 0xc0000a8c30, 0xc0000a8c00, 0x100cead, 0x1556e60, ...)
        /Users/andyw8/src/github.com/Shopify/themekit/src/httpify/client.go:105 +0x5d1
github.com/Shopify/themekit/src/httpify.(*HTTPClient).Put(0xc0002b2800, 0xc0002cba00, 0x33, 0x14ee800, 0xc0000a8c30, 0xc0000a8c00, 0xd, 0x19fa1e0, 0x0)
        /Users/andyw8/src/github.com/Shopify/themekit/src/httpify/client.go:82 +0x7f
github.com/Shopify/themekit/src/shopify.Client.UpdateAsset(0xc000337960, 0xc, 0xc0002b0ce0, 0x20, 0x19dd860, 0xd, 0xd, 0x19fa1e0, 0x0, 0x0, ...)
        /Users/andyw8/src/github.com/Shopify/themekit/src/shopify/theme_client.go:293 +0x1fd
github.com/Shopify/themekit/cmd.perform(0xc000020b40, 0xc0000f0bd0, 0x26, 0x0, 0xc000215b20, 0x20)
        /Users/andyw8/src/github.com/Shopify/themekit/cmd/watch.go:123 +0xa44
github.com/Shopify/themekit/cmd.watch(0xc000020b40, 0xc00015c900, 0xc00015c960, 0x167bf00, 0x19fa450, 0x0, 0x0)
        /Users/andyw8/src/github.com/Shopify/themekit/cmd/watch.go:86 +0x6d7
[...]

PR #791, opened 2020-09-11
Check that resp is not null before we check resp.StatusCode

topaztee:
fixes: [#790  ](https://github.com/Shopify/themekit/issues/790)


tanema:
Thanks @topaztee! Can you also add a test to make sure that this is an appropriate fix?

PR #806, opened 2020-09-23
Update template to prevent version confusion

andyw8:
@paulomarg and I have both misread the example version number in tickets, let's change this to prevent that happening again.

### Warn Checklist


paulomarg (approved):
This is fine with me, but I wonder if saying `e.g. running theme -v` or some such would work better (and tip people on how to get it).

andyw8:
@paulomarg good idea, updated.

paulomarg:
That's perfect, and even though I suggested the ` I think we might not want to add them - people might copy them and the command won't work as expected, we're even actively dropping those in other repos. Sorry!

PR #813, opened 2020-09-29
Extract RELEASING.md doc

andyw8:
Since the release instructions are aimed at a smaller audience than the contribution docs, I think it would be helpful to separate these.


RELEASING.md
```diff
@@ -0,0 +1,30 @@
+# Releasing Theme Kit
```
andyw8, line 1: I change the title here but everything else is the same.

tanema (approved):

PR #819, opened 2020-10-06
Http Client tweaks to idle conns to prevent hangs in connections

### LINKED ISSUE #801: theme download gets stuck at 97%
Hi there,

I recently updated themekit to latest version (1.1.1) and since then it cannot download all the files...
I tried with 3 different themes and all of them get stuck around 97-98%

`[development] Warning, this is the live theme on XXXXX-THEME.`
`[development] 389|401 [===================================================================>--]  97 %`

Any thoughts on this?

Thanks

**OS**: Mac
**Themekit**: v1.1.1

tanema:
will-fix #801 

With my experience doing highly concurrent http request in Go, I know that the default client has issues managing idle connection management, especially if it has no bounds on growth. I believe that this may be the issue that is causing the requests to hang. I debugged the hanging issues and the hang was actually happening on the request.

@MuhammadFarag I pinged you because you made the last tweaks to CH http client

### Warn Checklist

### DIFF
--- a/src/httpify/client.go
+++ b/src/httpify/client.go
@@ -2,11 +2,13 @@ package httpify
 
 import (
 	"bytes"
+	"context"
 	"crypto/tls"
 	"encoding/json"
 	"errors"
 	"fmt"
 	"io/ioutil"
+	"net"
 	"net/http"
 	"net/url"
 	"runtime"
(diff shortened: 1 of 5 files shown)


andyw8 (approved):

src/httpify/client.go
```diff
@@ -129,38 +129,46 @@ func (client *HTTPClient) doWithRetry(req *http.Request, body interface{}) (*htt
...
+		KeepAlive: 30 * time.Second,
 	}
 
 	return &http.Transport{
-		Proxy:           http.ProxyURL(parsedURL),
-		TLSClientConfig: &tls.Config{InsecureSkipVerify: true},
+		Proxy:                 proxy,
+		TLSClientConfig:       &tls.Config{InsecureSkipVerify: true},
+		IdleConnTimeout:       10 * time.Second,
+		TLSHandshakeTimeout:   time.Second,
```
andyw8, line 168: Very minor, but we could write `1  * time.Second` to keep it consistent with the other entries.

tanema:
Okay update, I also moved the http client into a single static variable. This is also recommended practice but also it is a lot better to have it as a static variable instead of wrapping it in a generator. I am not sure that this will help with performance or anything but it should make the http client configuration easier.

I have noticed that I no longer get hangs in downloading and it does seem like it uses slightly less cpu and memory but not really noticable, however I wonder if it would be more noticeable over time using watch.

src/httpify/client.go
```diff
@@ -117,51 +153,22 @@ func (client *HTTPClient) doWithRetry(req *http.Request, body interface{}) (*htt
...
 		client.limit.Wait()
-		resp, err := client.client.Do(req)
+		resp, err := httpClient.Do(req)
 		if err == nil {
 			if resp.StatusCode >= 100 && resp.StatusCode <= 428 {
 				return resp, nil
 			} else if resp.StatusCode == http.StatusTooManyRequests {
 				after, _ := strconv.ParseFloat(resp.Header.Get("Retry-After"), 10)
-				client.limit.ResetAfter(time.Duration(after))
+				client.limit.ResetAfter(time.Duration(after) * time.Second)
```
tanema, line 162: This is a critical fix. Before this the throttler was basically not working.

andyw8 (approved):

PR #827, opened 2020-10-14
Run gofmt

andyw8:
~~Once this is in, I'd like to add something a check to CI to ensure that we don't merge misformatted files.~~ Done in here too.


andyw8:
I skipped the `generated-assets` files, as running `make` after `gofmt` seems to cause them to change.

tanema (approved):

PR #832, opened 2020-10-22
Add note about Homebrew for Linux

andyw8:
The Homebrew distribution only includes a Darwin (Mac) binary so won't work on Linux.

### DIFF
--- a/docs/index.md
+++ b/docs/index.md
@@ -18,12 +18,14 @@ Using Theme Kit will enable you to
 
 ### macOS Installation
 
-Use [homebrew](http://brew.sh/) to install Theme Kit by running the following commands.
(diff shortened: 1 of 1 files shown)


andyw8:
(@tanema is there anything that would prevent distributing Theme Kit as source through Homebrew?).

tanema (approved):

tanema:
It would just make installation a lot larger, it would require adding Go as a dependency and then building themekit

paulomarg:
I think it might be possible to detect the OS running Homebrew and change the executable to download based on that.

andyw8:
Here's how Hugo does it, binaries for Mac then falls back to building with go:

https://github.com/Homebrew/homebrew-core/blob/master/Formula/hugo.rb

PR #839, opened 2020-11-18
Smarter Throttle Gate

tanema:
related #801 

This changes our rate limiter into a smart circuit breaking throttle gate. What this means functionally that:
- The request rate is limited to 4 per/second with bursts of 4 requests allowed
- When we get a 429 response, we cancel all inflight requests and they are all paused for the Retry-After value
- the allotted time is waited on
- The cancelled requests are started again
- The original request with the 429 response is restarted.
- None of these retries are counted against the http retry mechanism as they are not an error response from shopify or an error in the asset.

### Todo
Tests still need to be updated a little bit.

### Warn Checklist

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,6 +1,6 @@
 language: go
 go:
-  - 1.12
+  - 1.13
 script:
   - ./check_format.sh
   - env GO111MODULE=on make lint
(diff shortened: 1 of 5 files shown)


andyw8:
> None of these retries are not counted

Should that say "None of these retries are counted"?

andyw8:
> The request rate is limited to 4 per/second with bursts of 4 requests allowed

I think that's only for Plus, standard is 2 per second: https://shopify.dev/concepts/about-apis/rate-limits#rate-limiting-methods

tanema:
I am referring to themekit and not shopify. Since themekit doesnt really have a way to know what the merchant is, I have set it to the larger limit, and rely on the throttling to handle it. Since the range is not the different, non-plus people will not notice the difference.

Issue #631, opened 2019-04-24
Receiving malformed response (bad response or corrupted JSON)


PaulNewton:
**Is your feature request related to a problem? Please describe.**
Using version 1.0.0 windows 10 64 bit
When downloading theme files, using themekit get -t=THEMEID, you may get the following message output:
 `error downloading asset: received a malformed response`
Which seems to be from [ErrMalformedResponse](https://github.com/Shopify/themekit/blob/d6ee6553f6f5e88f352e953a4c66ed5384c83aad/src/shopify/theme_client.go#L26)
Without the verbose (-v) flag after the download process finishes this may be displayed multiple times.
![image](https://user-images.githubusercontent.com/47228209/56684520-ef2b4600-669d-11e9-8e1a-74a6a29d9b9d.png)
And the progress count indicates all files have downloaded but some files are actually missing when checking the folders&files on the OS.

With the verbose flag I will see the above error in between messages of "Successfully wrote X to disk" so creates an ambiguity because what does that mean when following an error?

-  A) the file displayed after the error message was the one erroring and finally downloaded successfully?
-  B) that a file got skipped **silently** and next message are other files downloading and wrote to disk successfully?
- ???

Without the filename or a checksum system you have to download the theme zip and do a comparison if you want to soothe the file paranoia, and then copy any missing files

**Describe the solution you'd like**
Included the relevant filename so interpretation of the message is easier.
Or clarifying if a file needs to be redownloaded either in the console or in the documentation.
Or provide a log of affected files to then pipe back into themekit for individual download.
Or prompt user to retry missing files

**Describe alternatives you've considered**
Alternative is to manually download theme zip from the store admin via email kinda rendering themekits download capability pointless when these errors occur.

**Additional context**
I'm not able to reliably reproduce the same output with the -v flag or narrow down if it's specific file types, or as a problem on shopifys end or my end.
The theme was created less than an hour before running command
Increasing the timeout to 1minute doesn't make it go away.

tanema:
Hey @PaulNewton thanks for opening the issue!

#### The error message
A malformed response should not be returned in normal usages, I will work on making that error message more informative however you should not be receiving that message. (Which is probably why that error message sucks in the first place)

#### Deeper debugging
I would love it if you could download a free copy of [Charles Proxy](https://www.charlesproxy.com/latest-release/download.do) When you run it the first time it should start recording with a proxy running on port 8888. 

[...]

PR #840, opened 2020-11-19
Adding more feedback while unmarshalling responses

tanema:
Related to #631

### Why is this PR needed

Right now, while unmarshalling responses, if the response is not JSON, then themekit will return an unhelpful error of `malformed response` While themekit does not really know what happened, and we have no way of knowing, we can add some more information to help analyze the issue.

### What this PR does

To the main `unmarshalResponse` method that the theme client uses, this adds
- http response status codes to the error messages. This should give us more insight into what the actual response is (500+ status codes)
- flushes the response body to a tempfile, so that it can be included in issue debugging. 
- outputs a path to the tempfile in the error message if flushing was successful.

### Warn Checklist
- [ ] ~I have :tophat:'d these changes by using the commands I changed by hand.~ it's hard to produce this error.

### DIFF
--- a/src/shopify/theme_client.go
+++ b/src/shopify/theme_client.go
@@ -370,7 +370,7 @@ func unmarshalResponse(resp *http.Response, data interface{}) error {
 	}
 	var re reqErr
 	mainErr := json.Unmarshal(reqBody, data) // check if we can unmarshal into the expected returned data
-	basicErr := json.Unmarshal(reqBody, &re) // if not returned data, check if we can get errors from the body
+	basicErr := json.Unmarshal(reqBody, &re) // if no returned data, check if we can get errors from the body
 	if mainErr != nil && basicErr != nil {
 		errStr := "could not unmarshal JSON from response body on a response with HTTP status %v. This usually means themekit received an html error page."
 		tmpFile, err := ioutil.TempFile(os.TempDir(), "themekit-response-*.txt")
--- a/src/shopify/theme_client.go
+++ b/src/shopify/theme_client.go
@@ -372,7 +372,7 @@ func unmarshalResponse(resp *http.Response, data interface{}) error {
 	mainErr := json.Unmarshal(reqBody, data) // check if we can unmarshal into the expected returned data
 	basicErr := json.Unmarshal(reqBody, &re) // if no returned data, check if we can get errors from the body
 	if mainErr != nil && basicErr != nil {
-		errStr := "could not unmarshal JSON from response body on a response with HTTP status %v. This usually means themekit received an html error page."
+		errStr := "could not unmarshal JSON from response body on a response with HTTP status %v. This usually means Theme Kit received an HTML error page."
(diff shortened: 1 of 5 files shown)


src/shopify/theme_client.go
```diff
@@ -361,20 +359,28 @@ func (c Client) assetPath(query map[string]string) string {
...
+	mainErr := json.Unmarshal(reqBody, data) // check if we can unmarshal into the expected returned data
+	basicErr := json.Unmarshal(reqBody, &re) // if not returned data, check if we can get errors from the body
 	if mainErr != nil && basicErr != nil {
-		return ErrMalformedResponse
+		errStr := "could not unmarshal JSON from response body on a response with HTTP status %v. This usually means themekit received an html error page."
+		tmpFile, err := ioutil.TempFile(os.TempDir(), "themekit-response-*.txt")
+		if err == nil {
+			tmpFile.Write([]byte(reqBody))
+			tmpFile.Close()
+			errStr += " Please find the full response at %v and include it with your ticket on github."
```
andyw8, line 380: Might this contain semi-sensitive data such as store names? We could suggest the people can edit it before posting?

andyw8:
Would it help to also include the Request ID in the error output?

src/shopify/theme_client.go
```diff
@@ -361,20 +359,28 @@ func (c Client) assetPath(query map[string]string) string {
...
-	err = body.Close()
+	err = resp.Body.Close()
 	if err != nil {
 		return err
 	}
 	var re reqErr
-	mainErr := json.Unmarshal(reqBody, data)
-	basicErr := json.Unmarshal(reqBody, &re)
+	mainErr := json.Unmarshal(reqBody, data) // check if we can unmarshal into the expected returned data
+	basicErr := json.Unmarshal(reqBody, &re) // if not returned data, check if we can get errors from the body
```
andyw8, line 373:
```suggestion
	basicErr := json.Unmarshal(reqBody, &re) // if no returned data, check if we can get errors from the body
```

src/shopify/theme_client.go
```diff
@@ -361,20 +359,28 @@ func (c Client) assetPath(query map[string]string) string {
...
 	if mainErr != nil && basicErr != nil {
-		return ErrMalformedResponse
+		errStr := "could not unmarshal JSON from response body on a response with HTTP status %v. This usually means themekit received an html error page."
```
andyw8, line 375:
```suggestion
		errStr := "could not unmarshal JSON from response body on a response with HTTP status %v. This usually means Theme Kit received an HTML error page."
```

src/shopify/theme_client.go
andyw8, line 380:
```suggestion
			errStr += " Please find the full response at %v and include it with your ticket on GitHub."
```

andyw8 (approved):

src/shopify/theme_client.go
tanema: No, generally these error pages will have only errors in them. They *might* have shop domain in them but that is not considered sensitive.

PR #844, opened 2020-11-19
Allowing python 3 usage

### LINKED ISSUE #725: Update python installation script to use Python 3.x
I want to install theme kit in a python docker environment, but I want to use Python 3.x. It would be great to have a Python script for this.

It would probably be a good solution, to create a separate installation script from the current one. This way installation scripts of others won't be impacted.

tanema:
fixes #725

The installation script only allowed python2 and not python3 but this simple change allows both versions for installation.

Works for both python2 and python3:
image

### Warn Checklist

### DIFF
--- a/docs/scripts/install.py
+++ b/docs/scripts/install.py
@@ -7,7 +7,7 @@
     Python Version: 2.7
     Description: Install script for themekit. It will download a release and make it executable
 '''
-import os, urllib, json, sys, hashlib
+import os, json, sys, hashlib
 
 class Installer(object):
     LATEST_RELEASE_URL = "https://shopify-themekit.s3.amazonaws.com/releases/latest.json"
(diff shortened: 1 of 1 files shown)


andyw8:
Is it worth trying to test the installation script as part of CI?

docs/scripts/install.py
```diff
@@ -25,7 +25,7 @@ def __init__(self, path="/usr/local/bin"):
         self.bin_path = "%s/theme" % self.install_path
         self.arch = self.__getArch()
         print("Fetching release data")
-        self.release = json.loads(urllib.urlopen(Installer.LATEST_RELEASE_URL).read().decode("utf-8"))
+        self.release = json.loads(self.__req(Installer.LATEST_RELEASE_URL).decode("utf-8"))
```
mllemango, line 28: Would it be any better to use `from six.moves.urllib.request import urlopen` instead of `self.__req`?

tanema: Will that work for both? Also what is `six.moves`

tanema: Ah I found out! https://six.readthedocs.io/ This is great! Thanks @mllemango !!

mllemango: Yup! Works on my end. 
https://six.readthedocs.io/#module-six.moves.urllib.request `six.moves.urllib.request` wraps py2's urllib and py3's urllib.request. Prevents you from having to write that extra module.

mllemango (approved):
So cool :D

PR #863, opened 2020-12-03
Remove six lib from install script

### LINKED ISSUE #856: six.moves.urllib.request added to install.py but no requirements.txt
Some of my deployment automation failed because `six.moves.urllib.request` was added in [this commit](https://github.com/Shopify/themekit/commit/b5acb8d3ae63758574652f719757a23307907509). Before you could run `install.py` with just the base Python and no dependencies. Could you publish a `requirements.txt` in case more dependencies are added?

(this is not urgent for me anyways right now because I am able to see the error and add the library manually myself).

tanema:
fixes #856 

The past change to the install script was cleaner but I thought `six.moves` was a stdlib and it turns out that is incorrect and the script should not have any deps, so I have changed back to this ugly version that does the version switch itself.

### Warn Checklist

### DIFF
--- a/docs/scripts/install.py
+++ b/docs/scripts/install.py
@@ -1,15 +1,13 @@
 #!/usr/bin/env python
 '''
-    File name: install.py
+    File name: install
     Author: Tim Anema
     Date created: Sep 29, 2016
-    Date last modified: Nov 19 2020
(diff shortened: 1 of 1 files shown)


mllemango:
😢

PR #875, opened 2020-12-14
CPU usage optimization in watch command

### LINKED ISSUE #664: Large CPU consumption with theme watch
**Describe the bug**
When running theme watch for a project with the following ignore_files setting in config.yml (no other settings aside from essential)..

```yaml
  ignore_files:
    - config/*
    - notes/*
    - node_modules/*
    - ./package.json
    - ./package.lock.json
    - prepros*
    - ./.vscode/*
    - liquid-testing/*
    - local-files/*
    - ./README.md
    - tests/*
    - src/*
    - webpack*.js
    - ./tsconfig.json
```
CPU usage starts and remains at at-least 15% - remaining between here and 21% - ish. I am using a fairly fast machine - see specs below.

Is is worth noting that the total project folder size is 141MB. 128MB belongs in the node_modules folder (which is ignored), and a lot of the remaining 13MB is also ignored in the other ignored directories. This only leaves a few MB of project files left that are being watched.

**To Reproduce**
1. Run command 'theme watch'.

**Expected behavior**
A low CPU usage, as is typically reported.

**Environment (please complete the following information):**
 - OS [e.g. iOS]: WIndows 10 (latest developers build)
- Themekit version [e.g. v0.7.5]: 1.0.1
 - Editor [e.g. atom, sublime]: VS Code 1.38.1
 - Machine Specs
   - Processor: Processor	Intel(R) Core(TM) i7-8565U CPU @ 1.80GHz, 1992 Mhz, 4 Core(s), 8 Logical Processor(s)
   - RAM: 16GB DDR4 2200MHz

tanema:
fixes #664

In previous versions, I watched each individual project directory however that created problems with creating new directories while the command run, and they were not tracked. Next step I took was to just watch all recursively and filter the events to just the project directories. This worked except, those with node_modules folders found that this caused their laptop to melt. 

This version now watched the root directory and project directories and does not filter out directory events. Now when a new directory is created, it will catch that event and start watching it, any deleted directories are discarded.

This should clear up the CPU issues.

### Bonus
I updated the dependency which allowed me to not have to parse new and old paths, and I am happy whenever I get to delete string parsing code.

### Warn Checklist

### DIFF
--- a/src/file/watcher.go
+++ b/src/file/watcher.go
@@ -6,7 +6,6 @@ import (
 	"io"
 	"os"
 	"path/filepath"
-	"strings"
 	"time"
 
 	"github.com/Shopify/themekit/src/env"
@@ -25,7 +24,6 @@ const (
 	Skip
 	// Get is when a file should be re-fetched, used in download operations
 	Get
-	filepathSplit = " -> "
 )
 
 var (
@@ -67,8 +65,14 @@ func NewWatcher(e *env.Env, configPath string, checksums map[string]string) (*Wa
 	}
 	fsWatcher.AddFilterHook(hook)
 
-	if err := fsWatcher.AddRecursive(e.Directory); err != nil {
+	if err := fsWatcher.Add(e.Directory); err != nil {
 		return nil, fmt.Errorf("Could not watch directory: %s", err)
+	}
+	for _, folder := range assetLocations {
+		path := filepath.Join(e.Directory, folder)
+		if err := fsWatcher.Add(path); err != nil && !os.IsNotExist(err) {
+			return nil, fmt.Errorf("Could not watch directory %s: %s", path, err)
+		}
 	}
 
 	return &Watcher{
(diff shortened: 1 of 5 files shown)


src/file/watcher.go
```diff
@@ -67,9 +65,15 @@ func NewWatcher(e *env.Env, configPath string, checksums map[string]string) (*Wa
 	}
 	fsWatcher.AddFilterHook(hook)
 
-	if err := fsWatcher.AddRecursive(e.Directory); err != nil {
+	if err := fsWatcher.Add(e.Directory); err != nil {
 		return nil, fmt.Errorf("Could not watch directory: %s", err)
 	}
+	for _, folder := range assetLocations {
+		path := filepath.Join(e.Directory, folder)
+		if err := fsWatcher.Add(path); err != nil && !os.IsNotExist(err) {
```
andyw8, line 73: Should this be `IsNotExist(path)`?

tanema: no, this is a helper that checks if the error is an error about it not existing https://golang.org/pkg/os/#IsNotExist The type system would not allow me to pass the wrong thing to it.

andyw8 (approved):

PR #889, opened 2021-01-25
Fixing folder events in file watcher

### LINKED ISSUE #885: Templates directory is trying to be located in a doubled-up absolute path
**Describe the bug**
AKA it's trying to process the absolute path of templates while inside the project directory.
It's happening while theme watch runs and I'm making changes to a template liquid file.

image

tanema:
fixes #885

### why

The file watcher would only originally catch directory events if they were the first event before the event buffer was started. This was because normally the event buffer is used to catch duplicate events or quick events like swap files which move and write at the same time. However there are some editors that touch the directory along with the file that was edited so the file edit event would come first, and the mod of the directory second, causing themekit to miss that event and it getting flushed through the rest of the system. 

### Fix

The directory event handling has been moved to the event translation step so that it will convert directory events into noops. This is a more central place to do this.

### Warn Checklist

### DIFF
--- a/src/file/watcher.go
+++ b/src/file/watcher.go
@@ -122,16 +122,14 @@ func (w *Watcher) watchFsEvents() {
 }
 
 func (w *Watcher) onEvent(event watcher.Event) bool {
-	if event.IsDir() {
-		if isEventType(event.Op, watcher.Create) {
-			w.fsWatcher.Add(event.Path)
-		}
+	events := map[string]Event{}
+	for _, e := range w.translateEvent(event) {
+		events[e.Path] = e
+	}
+	if len(events) == 0 {
 		return false
 	}
-	events := map[string]Event{}
-	for _, event := range w.translateEvent(event) {
-		events[event.Path] = event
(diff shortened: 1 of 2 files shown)


src/file/watcher.go
```diff
@@ -122,16 +122,14 @@ func (w *Watcher) watchFsEvents() {
...
+	for _, e := range w.translateEvent(event) {
+		events[e.Path] = e
 	}
+	if len(events) == 0 {
+		return false
+	}
+
 	drainTimer := time.NewTimer(drainTimeout)
 	defer drainTimer.Stop()
 	for {
```
tanema, line 135: The actual issue is on line [143](https://github.com/Shopify/themekit/pull/889/files#diff-a281c8cf40c688f804f3d1a0a98412d64de22cc399d2aae1682d2861bb5cf467L139-L146) where we get a second stage of events, the directory event filtering from line 125 was not duplicated there.

andyw8 (approved):

